PR #9, opened 2021-03-19
Key generation code

lukehinds:
This PR implements key generation with algorithm agility (in case
different projects have certain crypto compliance requirements)

After this PR is merged, the ODIC call will be pushed.


cmd/pkg/keymgmt/keymgmt.go
```diff
@@ -0,0 +1,67 @@
...
+	"crypto/rsa"
+	"errors"
+)
+
+func GeneratePrivateKey(algorithm string) (interface{}, interface{}, error) {
+	var err error = nil
+	var key interface{}
+	// Allow algorithm agility if different projects have certain FIPS like compliance requirements
+	switch algorithm {
+	case "1024":
```
dlorenc, line 31: I think we can drop 1024, it's considered broken.

cmd/pkg/keymgmt/keymgmt.go
```diff
@@ -0,0 +1,67 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"errors"
+)
+
+func GeneratePrivateKey(algorithm string) (interface{}, interface{}, error) {
+	var err error = nil
+	var key interface{}
+	// Allow algorithm agility if different projects have certain FIPS like compliance requirements
+	switch algorithm {
```
dlorenc, line 30: we probably need to include RSA in the prefix here, same as below with ecdsa.

cmd/pkg/keymgmt/keymgmt.go
```diff
@@ -0,0 +1,67 @@
...
+	case "3072":
+		key, err = rsa.GenerateKey(rand.Reader, 3072)
+	case "P224":
+		key, err = ecdsa.GenerateKey(elliptic.P224(), rand.Reader)
+	case "P256":
+		key, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	case "P384":
+		key, err = ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
+	case "P521":
+		key, err = ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
```
dlorenc, line 44: I'd also be in favor of adding ed25519 if we're going to add all of these :)

cmd/pkg/keymgmt/keymgmt.go
lukehinds: agree, I put it there in case something obscure may have needed it for a poc with their current old system, but its better out.

cmd/pkg/keymgmt/keymgmt.go
lukehinds: That curve is not in the library (unless I am looking in the wrong place): 

https://golang.org/pkg/crypto/elliptic/

lukehinds: ok, got you https://golang.org/pkg/crypto/ed25519/

do you mind if I do that in a follow up PR?

dlorenc: Sure!

dlorenc: https://golang.org/pkg/crypto/ed25519/ for reference - it's a separate package.

cmd/pkg/keymgmt/keymgmt.go
dlorenc: Thanks - definitely a balance between supporting things people need and perpetuating known broken stuff. I think this one is a clear case of being broken enough to drop it.

dlorenc (approved):

PR #13, opened 2021-03-20
Return PubK in correct type

lukehinds:
PubK requires MarshalPKIXPublicKey and returning as []uint8

### DIFF
--- a/pkg/keymgmt/keymgmt.go
+++ b/pkg/keymgmt/keymgmt.go
@@ -20,18 +20,15 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/x509"
 	"errors"
 )
 
-func GeneratePrivateKey(algorithm string) (interface{}, interface{}, error) {
+func GeneratePrivateKey(algorithm string) (interface{}, []uint8, error) {
 	var err error
 	var key interface{}
 	// Allow algorithm agility if different projects have certain FIPS like compliance requirements
 	switch algorithm {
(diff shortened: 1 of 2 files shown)


pkg/keymgmt/keymgmt.go
```diff
@@ -43,31 +44,36 @@ func GeneratePrivateKey(algorithm string) (interface{}, interface{}, error) {
...
 		pub = &k.PublicKey
 	case *ecdsa.PrivateKey:
 		pub = &k.PublicKey
-	// The below would never happen (unless upstream crypto is broken, but lets log it anyway)
 	default:
 		err = errors.New("error generating public key" )
 	}
+	if err != nil {
+		return nil, err
+	}
```
dlorenc, line 72: in some of the cosign code I actually gave up with this error and threw a panic (same as below during MarshalPKIXPublicKey). Like you said, an examination of the code shows no error can be thrown here. It really simplifies callers to not have to deal with this error. A panic seems appropriate. Your call though!

lukehinds: I tend to agree with that, if the underlying code breaks there is really no recovering and its appropriate to panic.

dlorenc:
Lgtm!

lukehinds:
Thanks @dlorenc , I forget to mention, I dropped RSA for now, as we don't handle it server side

https://github.com/sigstore/fulcio/blob/dbc10540245a5484c5f65419bb61c83b9e34aa22/pkg/ca/ca.go#L50-L60

PR #14, opened 2021-03-22
Client port

lukehinds:


pkg/oauthflow/pkce.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	var chosenMethod string
+	for _, method := range providerClaims.CodeChallengeMethodsSupported {
+		// per RFC7636, any server that supports PKCE must support S256
+		if method == PKCES256 && chosenMethod != PKCES256 {
+			chosenMethod = PKCES256
+			break
+		} else if method != "plain" {
+			log.Logger.Infof("Unsupported code challenge method in list: '%v'", method)
```
dlorenc, line 56: I know you just copied this over, but we might want to use the cli.Logger here rather than the production one.

dlorenc (approved):
Approved, just a few nits that would be fine to handle later with TODOs.

pkg/oauthflow/pkce.go
lukehinds: yup, totally. makes sense

bobcallaway:
Is there a `Makefile` that can be reused to generate the client stubs?

lukehinds:
> Is there a `Makefile` that can be reused to generate the client stubs?

Yes, I did plan to do that. I was going to talk with you both about how we share the openapi.yml from fulcio into here.

lukehinds:
> Is there a `Makefile` that can be reused to generate the client stubs?

I wanted to chat to you about this, as in how we handle the openapi.yml in this repo. I thought we could either:

Could we do this as a follow up patch? Makes sense to do this as the same time as docker compose files.

Makefile
```diff
@@ -0,0 +1,32 @@
+.PHONY: all test clean lint gosec
+
+all: client
```
lukehinds, line 3: I guess we don't need this, as there is no 'server' for constituent an 'all'

PR #19, opened 2021-03-27
Rename httpclient

lukehinds:


lukehinds:
This is branched on top of https://github.com/sigstore/sigstore/pull/18 , so that can be

PR #21, opened 2021-03-27
Implements file MIME checking

lukehinds:
This will be useful for either checking it is a supported type
or using as a method to take specific actions on certain
mime types.

I did start to write this myself, but the upstream library
is really decent and well benchmarked.

### DIFF
--- a/cmd/sign.go
+++ b/cmd/sign.go
@@ -23,9 +23,10 @@ import (
 	"encoding/pem"
 	"fmt"
 	"github.com/sigstore/sigstore/pkg/generated/client/operations"
-	"github.com/sigstore/sigstore/pkg/httpclient"
+	"github.com/sigstore/sigstore/pkg/httpclients"
 	"github.com/sigstore/sigstore/pkg/keymgmt"
 	"github.com/sigstore/sigstore/pkg/oauthflow"
+	"github.com/sigstore/sigstore/pkg/utils"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"net/http"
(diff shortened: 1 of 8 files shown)


pkg/utils/filehandler.go
```diff
@@ -0,0 +1,26 @@
+package utils
+
+import (
+	"github.com/gabriel-vasile/mimetype"
+)
+
+var SupportedFileTypes = []string{
```
dlorenc, line 7: I think a common pattern here is to use a map[string]struct{}{} as a set, rather than needing to iterate through a list.

cmd/sign.go
```diff
@@ -88,6 +89,16 @@ var signCmd = &cobra.Command{
 			panic("failed to parse certificate: " + err.Error())
 		}
 		fmt.Printf("Received signing Cerificate: %+v\n", cert.Subject)
+
+		mimetype, err := utils.GetFileType(viper.GetString("artifact"))
+		if err != nil {
+			fmt.Println(err)
+		}
+
+		_, found := utils.FindString(utils.SupportedFileTypes, mimetype)
```
dlorenc, line 98: nit: inline this if statement

dlorenc (approved):

cmd/sign.go
```diff
@@ -94,7 +95,17 @@ var signCmd = &cobra.Command{
 			panic("failed to parse certificate: " + err.Error())
 		}
 		fmt.Printf("Received signing Cerificate: %+v\n", cert.Subject)
-		},
+
+		mimetype, err := utils.GetFileType(viper.GetString("artifact"))
+		if err != nil {
+			fmt.Println(err)
```
dlorenc, line 101: we should probably return or panic here instead of continuing? same as on 106

lukehinds: Good point, I will sys exit with an error exit code (as its a CLI), sound good @dlorenc ?

PR #23, opened 2021-04-01
Implement rekor log entry

lukehinds:


pkg/signfile/signfile.go
```diff
@@ -0,0 +1,80 @@
...
+		return "", err
+	}
+	// UUID is at the end of location
+	for _, p := range resp.Payload {
+		return strconv.FormatInt(*p.LogIndex, 10), nil
+	}
+	return "", errors.New("bad response from server")
+}
+
+func MarshalPublicKey(pub *ecdsa.PublicKey) ([]byte, error) {
```
dlorenc, line 67: I'd still love to try to turn these errors into panics, they should never happen

cmd/sign.go
```diff
@@ -53,18 +55,26 @@ var signCmd = &cobra.Command{
...
 			fmt.Println(err)
 		}
 
-		h := sha256.Sum256([]byte(email))
-		proof, err := ecdsa.SignASN1(rand.Reader, key.(*ecdsa.PrivateKey), h[:])
+		privKey := key.(*ecdsa.PrivateKey)
+		pubKey := pub.(*ecdsa.PublicKey)
+		pubBytes, err := x509.MarshalPKIXPublicKey(pubKey)
+		if err != nil {
+			fmt.Println(err)
```
dlorenc, line 68: should return here and above on 61

cmd/sign.go
```diff
@@ -105,14 +115,40 @@ var signCmd = &cobra.Command{
 		if !result {
 			fmt.Printf("File type %s currently not supported\n", mimetype)
 		}
+
+		// Hash the artifact
+		payload, err := ioutil.ReadFile(viper.GetString("artifact"))
+		if err != nil {
+			fmt.Println("Error reading file {}", viper.GetString("artifact"))
```
dlorenc, line 122: should return/panic

cmd/sign.go
lukehinds: I don't think anything we can return to anything as this is called from &cobra.Command, what would we return to, root.go?

pkg/signfile/signfile.go
dekkagaijin: they will be able to happen if/when we accept `crypto.PublicKey` instead of a known-good concrete type

cmd/sign.go
dlorenc: Ah, you can use the RunE method in cobra so you can return an error

lukehinds: Got you, I will look to implement this.

dlorenc (approved):

PR #24, opened 2021-04-01
Device flow!

dlorenc:
Signed-off-by: Dan Lorenc <dlorenc@google.com>


pkg/oauthflow/device.go
```diff
@@ -0,0 +1,146 @@
...
+func NewDeviceFlowTokenGetter(issuer string) *DeviceFlowTokenGetter {
+	return &DeviceFlowTokenGetter{
+		MessagePrinter: func(s string) { fmt.Println(s) },
+		Sleeper:        time.Sleep,
+		Issuer:         issuer,
+	}
+}
+
+func (d *DeviceFlowTokenGetter) deviceFlow(clientID string) (string, error) {
+	deviceUrl := d.Issuer + "/device/code"
```
dekkagaijin, line 60: Both request endpoints are arbitrary in the spec. Google uses a discovery endpoint to advertise which to use: https://developers.google.com/identity/protocols/oauth2/openid-connect#discovery

pkg/oauthflow/device.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", err
+	}
+	d.MessagePrinter(fmt.Sprintf("Open your browser to: %s", parsed.VerificationUriComplete))
+
+	pollUrl := d.Issuer + "/device/token"
+	for {
+		data := url.Values{
+			"grant_type":  []string{"urn:ietf:params:oauth:grant-type:device_code"},
+			"device_code": []string{parsed.DeviceCode},
+			"scope":       []string{"openid", "email"},
```
dekkagaijin, line 87: No `scope` in the token request: https://tools.ietf.org/html/rfc8628#section-3.4

Google does issue a `client_secret`, even if it isn't secret: https://developers.google.com/identity/protocols/oauth2/limited-input-device#step-4:-poll-googles-authorization-server

dlorenc: Ah, nice. I left a comment to track this, we don't use a secret with our federated provider

pkg/oauthflow/device.go
dlorenc: Nice catch. Fixed this one. The discovery stuff is too buried in the oauth2 library we use to be useful so I made it configurable.

pkg/oauthflow/device.go
```diff
@@ -0,0 +1,157 @@
...
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+	parsed := deviceResp{}
+	if err := json.Unmarshal(b, &parsed); err != nil {
+		return "", err
+	}
+	d.MessagePrinter(fmt.Sprintf("Open your browser to: %s", parsed.VerificationUriComplete))
```
bobcallaway, line 90: https://tools.ietf.org/html/rfc8628#section-3.2 states that the `verification_uri_complete` field is optional... so might be better to wrap this with a check and substitute `verification_uri` and `user_code` if its not set?

pkg/oauthflow/device.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+const (
+	/* #nosec */
+	SigstoreDeviceURL = "https://oauth2.sigstore.dev/auth/device/code"
+	/* #nosec */
+	SigstoreTokenURL = "https://oauth2.sigstore.dev/auth/device/token"
+)
+
+type deviceResp struct {
```
bobcallaway, line 39: might be useful to get the `expires_in` timestamp and print it out to the console

pkg/oauthflow/device.go
```diff
@@ -0,0 +1,157 @@
...
+		if tr.IDToken != "" {
+			d.MessagePrinter("Token received!")
+			return tr.IDToken, nil
+		}
+		switch tr.Error {
+		case "access_denied", "expired_token":
+			return "", fmt.Errorf("error obtaining token: %s", tr.Error)
+		case "authorization_pending":
+		case "slow_down":
+		default:
```
bobcallaway, line 123: note that the error codes from https://tools.ietf.org/html/rfc6749#section-5.2 are still valid and we would end up in a loop repeating when we shouldn't.

need to move Sleeper calls up to cover the last two cases but not this one,  since there is no implicit fall through in golang.

pkg/oauthflow/device.go
```diff
@@ -0,0 +1,157 @@
...
+
+		if tr.IDToken != "" {
+			d.MessagePrinter("Token received!")
+			return tr.IDToken, nil
+		}
+		switch tr.Error {
+		case "access_denied", "expired_token":
+			return "", fmt.Errorf("error obtaining token: %s", tr.Error)
+		case "authorization_pending":
+		case "slow_down":
```
bobcallaway, line 122: should do interval += 5 seconds here for slow_down

pkg/oauthflow/device.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return nil, err
+	}
+	verifier := p.Verifier(&oidc.Config{ClientID: cfg.ClientID})
+	parsedIDToken, err := verifier.Verify(context.Background(), idToken)
+	if err != nil {
+		return nil, err
+	}
+
+	var claims struct {
```
bobcallaway, line 142: can use `EmailFromIDToken` instead of duplicating the logic

lukehinds:
I don't know enough about device flow at present to make a meaningful review, but the code looks good to me.

pkg/oauthflow/device.go
dlorenc: Good catch!

pkg/oauthflow/device.go
dlorenc: Nice!

pkg/oauthflow/device.go
dlorenc: Good catch.

pkg/oauthflow/device.go
dlorenc: I think I got it.

pkg/oauthflow/device.go
dlorenc: EmailFromIDToken actually disappeared somehow. Added it back

bobcallaway (commented):
otherwise LGTM. good stuff!

cmd/sign.go
```diff
@@ -68,7 +70,9 @@ var signCmd = &cobra.Command{
 		idToken, email, err := oauthflow.OIDConnect(
 			viper.GetString("oidc-issuer"),
 			viper.GetString("oidc-client-id"),
-			viper.GetString("oidc-client-secret"))
+			viper.GetString("oidc-client-secret"),
+			oauthflow.DefaultIDTokenGetter,
```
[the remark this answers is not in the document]
lukehinds: I was about to ask the same. +1 to being able to select which mode to operate with.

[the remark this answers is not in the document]
dlorenc: Yeah, I'm trying to figure out the best way to do that. A flag probably works. In my WIP PR in cosign I was going to default to device if there's no terminal attached.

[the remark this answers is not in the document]
dlorenc: I'll open another issue to track this here.

PR #26, opened 2021-04-03
Add `signature` library

dekkagaijin:
The goal is it as easy as possible to put the business logic in `pkg`, re-use `pkg` across project within `sigstore`, and allow third parties to build on top of these libraries (e.g. to implement CI plugins)

TODOs: 
* refactor existing libs in `sigstore` and `cosign`
* implement TPM signer


lukehinds:
code looks good @dekkagaijin , it would be good to get some context on the main purpose of the PR. It looks like this would afford us more agility around switching algorithms (we can add more curves alongside `pkg/signature/ed25519.go`). 

In regards to `payload/payload.go` this looks quite specific to containers, my guess is only cosign would import and use this code?  

If you could drop in a couple of sentences on how this will be leveraged by the clients that would be great!

dlorenc:
> In regards to `payload/payload.go` this looks quite specific to containers, my guess is only cosign would import and use this code?

Ah yeah, payload.go the container util should probably stay in cosign.

lukehinds:
> > In regards to `payload/payload.go` this looks quite specific to containers, my guess is only cosign would import and use this code?
> 
> Ah yeah, payload.go the container util should probably stay in cosign.

I don't really mind its location, main point was that it did not become the default way of handling a payload. e.g. if it's modular that's fine by me.

dekkagaijin:
> Ah yeah, payload.go the container util should probably stay in cosign.

Isn't the goal of `sigstore` to be as broad as possible?

dekkagaijin:
> code looks good @dekkagaijin , it would be good to get some context on the main purpose of the PR. It looks like this would afford us more agility around switching algorithms (we can add more curves alongside `pkg/signature/ed25519.go`).
> 
> In regards to `payload/payload.go` this looks quite specific to containers, my guess is only cosign would import and use this code?
> 
> If you could drop in a couple of sentences on how this will be leveraged by the clients that would be great!

Basically I want to make it as easy as possible to put the business logic in `pkg`, re-use `pkg` across project within `sigstore`, and allow third parties to build on top of these libraries (e.g. to implement CI plugins). I'd been going through the same process within `cosign` as a preamble to implementing additional signers until I found out about `sigstore/sigstore`

dlorenc:
Good question actually. I don't know if it's intended to be a strict superset of the functionality in the other tools. 

I guess it's a question of broad vs. generic. In my head - anything container specific or OCI registry specific would live in cosign, while sigstore would deal with blobs and tarballs. But that doesn't really make a ton of sense now that I think about it.

dekkagaijin:
> I guess it's a question of broad vs. generic. 

Fair, makes sense. I think we can accomplish both, as long as we aim for generic and broad in that order

lukehinds:
Yep, the idea was to make this tool be a capture all for general signing of blobs, tarballs, sboms, exe's (I just made the last one up). It also made sense that instead of having several different tools all implement their own implementations of oidc, signing etc, we could maintain one set in here that can be used by all tools. So that said I am supportive of this PR and think its on the right path. The only thing I wanted to check was that container specifics were not going to be the default validations or canonicalization for all possible consumers of this clients pkgs (which it looks like they are not). Hope that makes sense.

dlorenc:
Yup! The only bit I'm not sure of is whether we imagine the container bits existing here too. Either way, fine for now and we can clean up later!

lukehinds:
I just noticed 

> implement TPM signer

That sounds interesting, what's the plan there?

dlorenc (approved):

dekkagaijin:
> I just noticed
> 
> > implement TPM signer
> 
> That sounds interesting, what's the plan there?

Basically a `Signer` that liberally leverages `go-tpm` and `go-tpm-tools` to sign payload digests.
I'm thinking we can significantly improve the UX for this use case and enable hardware trust assertions

lukehinds:
> > I just noticed
> > > implement TPM signer
> > 
> > 
> > That sounds interesting, what's the plan there?
> 
> Basically a `Signer` that liberally leverages `go-tpm` and `go-tpm-tools` to sign payload digests.
> I'm thinking we can significantly improve the UX for this use case and enable hardware trust assertions

Cool, it would be good to discuss this in an issue first. Can you create one and mark it as label it as pathfinding.

PR #35, opened 2021-04-09
Signers should return the payloads which were actually signed

dekkagaijin:


dlorenc (approved):
I like it!

PR #38, opened 2021-04-09
Leverage the `signature` package for signing

dekkagaijin:

### DIFF
--- a/cmd/sign.go
+++ b/cmd/sign.go
@@ -16,9 +16,7 @@
 package cmd
 
 import (
-	"crypto/ecdsa"
-	"crypto/rand"
-	"crypto/sha256"
+	"context"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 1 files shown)


lukehinds:
I will try and test this and review later today, just need to resolve some CA connection issues I have.

cmd/sign.go
```diff
@@ -78,20 +76,21 @@ var signCmd = &cobra.Command{
...
-		pubKey := pub.(*ecdsa.PublicKey)
-		pubBytes, err := x509.MarshalPKIXPublicKey(pubKey)
+		pubBytes, err := x509.MarshalPKIXPublicKey(pub)
 		if err != nil {
 			return err
 		}
 
-		emailHash := sha256.Sum256([]byte(email))
-		proof, err := ecdsa.SignASN1(rand.Reader, privKey, emailHash[:])
+		proof, _, err := signer.Sign(ctx, []byte(email))
```
lukehinds, line 93: Does signer.Sign also generate a hash as well as sign?

dlorenc: Yeah, my understanding is that it's "signature, signed, error", some algorithms (ed25519) do hashing on their own, others require you to prehash, some like fido2 actually mutate the input payload in other ways before signing.

lukehinds: ok, got it now.

lukehinds (approved):

cmd/sign.go
dekkagaijin: Yup, what Dan said.

PR #43, opened 2021-04-16
use RSA-PSS instead of RSA-PKCS#1 v1.5 signature scheme

dekkagaijin:
PKCS#1 v1.5 has known attacks

### DIFF
--- a/pkg/signature/rsa.go
+++ b/pkg/signature/rsa.go
@@ -25,8 +25,8 @@ import (
 )
 
 type RSAVerifier struct {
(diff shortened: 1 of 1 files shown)


dlorenc:
Is this actually used anywhere?

Either way, LGTM.

dlorenc (approved):

dekkagaijin:
> Is this actually used anywhere?

Not afaik, just want there to be a sane default.

dlorenc:
> Not afaik, just want there to be a sane default.

Thanks!

PR #45, opened 2021-04-24
Allow specifying the full key version.

dlorenc:
This means we don't need to give KMS key management IAM permissions
if you can specify the full thing.

Signed-off-by: Dan Lorenc <dlorenc@google.com>

### DIFF
--- a/pkg/kms/gcp/gcp.go
+++ b/pkg/kms/gcp/gcp.go
@@ -39,12 +39,13 @@ type KMS struct {
 	locationID    string
 	keyRing       string
 	key           string
+	version       string
 }
 
 var (
-	ErrKMSReference = errors.New("kms specification should be in the format gcpkms://projects/[PROJECT_ID]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]")
+	ErrKMSReference = errors.New("kms specification should be in the format gcpkms://projects/[PROJECT_ID]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]/versions/[VERSION]")
 
-	re = regexp.MustCompile(`^gcpkms://projects/([^/]+)/locations/([^/]+)/keyRings/([^/]+)/cryptoKeys/([^/]+)$`)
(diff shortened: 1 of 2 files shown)


dlorenc:
cc @priyawadhwa

priyawadhwa (dismissed):
nice!

priyawadhwa:
looks like i need write access to this repo 😅

pkg/kms/gcp/gcp.go
```diff
@@ -39,12 +39,13 @@ type KMS struct {
 	locationID    string
 	keyRing       string
 	key           string
+	version       string
 }
 
 var (
 	ErrKMSReference = errors.New("kms specification should be in the format gcpkms://projects/[PROJECT_ID]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]")
```
bobcallaway, line 46: do you need to update this string to note the optional versions tag?

pkg/kms/gcp/gcp.go
```diff
@@ -57,18 +58,18 @@ func ValidReference(ref string) error {
...
-func parseReference(resourceID string) (projectID, locationID, keyRing, keyName string, err error) {
+func parseReference(resourceID string) (projectID, locationID, keyRing, keyName, version string, err error) {
 	v := re.FindStringSubmatch(resourceID)
-	if len(v) != 5 {
+	if len(v) < 5 {
 		err = errors.Errorf("invalid gcpkms format %q", resourceID)
 		return
 	}
-	projectID, locationID, keyRing, keyName = v[1], v[2], v[3], v[4]
+	projectID, locationID, keyRing, keyName, version = v[1], v[2], v[3], v[4], v[5]
```
bobcallaway, line 67: if `len(v) = 5` then wouldn't `v[5]` be out of bounds?

bobcallaway (commented):
two questions

pkg/kms/gcp/gcp.go
dlorenc: Good catch.

bobcallaway (approved):
lgtm

PR #49, opened 2021-04-27
some vault fixes

RichiCoder1:
~I noted this extra write, but I'm not sure this will actually fix the issue. I couldn't reproduce the pub key issue via e2e.~

I updated the tests to use the same `ECDSAVerifier` that cosign uses and sure enough it started failing. I ended up shifting the hash to hash locally instead of using Vault's hash endpoint and it seems happy now.


test/e2e/kms/vault_test.go
```diff
@@ -104,6 +106,40 @@ func (suite *VaultSuite) TestSign() {
...
+	ok = ecdsa.VerifyASN1(pubKey, hash[:], sig)
+	assert.True(suite.T(), ok)
+
+	ok = ecdsa.VerifyASN1(key, hash[:], sig)
+	assert.True(suite.T(), ok)
+
+	verifier := signature.ECDSAVerifier{
+		Key:     pubKey,
+		HashAlg: crypto.SHA256,
+	}
```
RichiCoder1, line 139: In theory this is exactly what is happening in `cosign verify`. I wonder if something is getting lost in  translation.

RichiCoder1:
Unrelated: It looks like `pkg/oauthflow/flow_test.go` is experiencing a race condition

dlorenc (approved):
Wow, thanks for the debugging!

dlorenc:
> Unrelated: It looks like `pkg/oauthflow/flow_test.go` is experiencing a race condition

Opened an issue.

PR #51, opened 2021-05-03
Add support for ed25519 based keys

priyawadhwa:
We can use this in tekton chains! 

Signed-off-by: Priya Wadhwa <priyawadhwa@google.com>

### DIFF
--- a/pkg/signature/ed25519.go
+++ b/pkg/signature/ed25519.go
@@ -1,0 +1,71 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package signature
+
+import (
+	"context"
+	"crypto"
+	"crypto/ed25519"
+	"crypto/rand"
+	_ "crypto/sha256" // To ensure `crypto.SHA256` is implemented.
+	"errors"
+	"fmt"
+)
+
+type ED25519Verifier struct {
+	Key ed25519.PublicKey
+}
+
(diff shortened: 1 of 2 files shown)


lukehinds (commented):
~~how do you see this being called? I can't see any wiring up to the APIs. Is this instead something that will be imported into chains and used there?~~

forget the above, I thought I was in sigstore/fuclio

pkg/signature/ed25519.go
```diff
@@ -0,0 +1,71 @@
...
+
+type ED25519SignerVerifier struct {
+	ED25519Verifier
+	Key ed25519.PrivateKey
+}
+
+func (s ED25519SignerVerifier) Sign(_ context.Context, rawPayload []byte) (signature, signed []byte, err error) {
+	signature = ed25519.Sign(s.Key, rawPayload)
+	return
+}
```
lukehinds, line 40: might be worth capturing an err here, such as:

https://github.com/sigstore/sigstore/blob/11e6eaab7cdc5cea7113004d2de344fe0f7fb640/pkg/signature/ecdsa.go#L45-L48

lukehinds (commented):
The code looks good, but just thinking about the layout / organisation. 

We already have `/pkg/signature/ecdsa.go` could we not just pass in the curve type we want there and have all ecliptic handling in ecdsa.go 

e.g. use the below and instead of having `elliptic.P256()` set, we allow folks to pass in their own preferred code?

https://github.com/sigstore/sigstore/blob/11e6eaab7cdc5cea7113004d2de344fe0f7fb640/pkg/signature/ecdsa.go#L80-L86

pkg/signature/ed25519.go
priyawadhwa: The [ed25519.Sign](https://golang.org/pkg/crypto/ed25519/#Sign) function actually only returns `[]byte` :)

priyawadhwa:
>We already have /pkg/signature/ecdsa.go could we not just pass in the curve type we want there and have all ecliptic handling in ecdsa.go

IIUC ecdsa & ed25519 are different algorithms, so I don't think passing in a different curve to ecdsa would necessarily give you ed25519 (but I could be wrong about that)

dlorenc:
> IIUC ecdsa & ed25519 are different algorithms, so I don't think passing in a different curve to ecdsa would necessarily give you ed25519 (but I could be wrong about that)

Yup, different curves.

bobcallaway (approved):

PR #63, opened 2021-05-21
Move all rekor code to tlog

lukehinds:
The code in `signing.go` makes little sense, why not have it all
in tlog. I made this change while looking at leveraging the code
within the tektoncd/chains project and importing sigstore/signfile
looks confusing.


lukehinds:
anyone reviewing, I had to add a `//nolint` to ed2559.go which is from an earlier patch unrelated to this change (not sure why it never failed before, maybe the linter has been updated), anyway if looks like a false positive to me.

dlorenc:
> anyone reviewing, I had to add a `//nolint` to ed25519.go which is from an earlier patch unrelated to this change (not sure why it never failed before, maybe the linter has been updated), anyway if looks like a false positive to me.

I patched that in #64

dlorenc (approved):

PR #67, opened 2021-05-26
enable transit secret engine at another path

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

I have a problem with storing keys in the Vault transit engine. cosign waits for me to enable `Vault Transit Secret Engine` at path `transit/`, but in Vault, I don’t have to enable secret engines at its default paths, so, I can override this behavior by using -path option while enabling it. So, in the scenario I have been enabled Transit Secret Engine at a different path, cosign throws an error like the following:
Screen Shot 2021-05-26 at 20 32 19

if you want to reproduce that error above, you can type the following commands:
```
$ vault server -dev -dev-root-token-id root
$ export VAULT_ADDR="http://localhost:8200"
$ export VAULT_TOKEN="root"
$ vault secrets enable -path=keystore transit
$ cosign generate-key-pair -kms hashivault://test-key
```

This PR is aiming to solve this problem that I mentioned above.

### DIFF
--- a/pkg/kms/hashivault/hashivault.go
+++ b/pkg/kms/hashivault/hashivault.go
@@ -34,8 +34,9 @@ import (
 )
 
 type KMS struct {
-	client  *vault.Client
-	keyPath string
+	client                  *vault.Client
+	keyPath                 string
+	transitSecretEnginePath string
 }
 
 var (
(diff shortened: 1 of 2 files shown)


developer-guy:
cc: @dlorenc @Dentrax

dlorenc:
Hmm, maybe one idea - what if we put this in the "vault_addr" env var rather than adding a new one? If there's no "suffix", we can default to "/transit", otherwise we can just use the full value. Does that make sense?

developer-guy:
> Hmm, maybe one idea - what if we put this in the "vault_addr" env var rather than adding a new one? If there's no "suffix", we can default to "/transit", otherwise we can just use the full value. Does that make sense?

Hello @dlorenc, thank you for an idea but `VAULT_ADDR` environment variable is kind of a special öne for Vault, we should not change it. This is why I don't put `VAULT_` prefix to my environment variable to show that this is not a special environment variable that Vault needs.

dlorenc:
Ah! Makes sense :)

PR #69, opened 2021-06-03
convert signature library to implement crypto.Signer interface

bobcallaway:
This is the first step of several to try to have signature library that can be used across all of sigstore's golang projects.

1. Have all *non-KMS* implementations implement crypto.Signer interface
2. Support hash algorithms other than SHA256
3. Replaces calls of fmt.Print(f|ln) to log.Printf()
4. Adds public key caching support for KMS providers
5. Use correct hashing algorithm per KMS provider response


dlorenc:
cc @dekkagaijin

dekkagaijin:
The reason why I intentionally did _not_ implement the `crypto.Signer` interface is because it does not support plumbing through `Context`s for RPC-based implementations (e.g. KMS). What's the end goal?

dekkagaijin:
tl;dr `crypto.Signer` isn't an abstraction. Hopefully we can improve things by providing `sigstore.Signer` as a useful superset of that API

bobcallaway:
> The reason why I intentionally did _not_ implement the `crypto.Signer` interface is because it does not support plumbing through `Context`s for RPC-based implementations (e.g. KMS). 

Understood... by shifting the public key into a cache (which refreshes on its own context) and leveraging the `crypto.SignerOpts` interface to pass the context into the `Sign()` implementation, I think we could achieve this and still implement the interface. 

> Additionally, we also need to return the input to the underlying `crypto.Signer` in order to support things like `cosign` which want/need to know any transformations applied to the raw payload (e.g. digest). It's a higher-level interface to allow for the broadest possible implementation and plugability.
> 
> > Support hash algorithms other than SHA256
> > Replaces calls of fmt.Print(f|ln) to log.Printf()
> > Adds public key caching support for KMS providers
> > Use correct hashing algorithm per KMS provider response
> 
> I like these. Hash algs other than SHA256 are already supported via constructors, but can't sanely be used as an input to `Sign` and `Verify` since not all signing algs allow for pre-hashed inputs.

Yup, you're right - I should combine the first and fourth items above since I really meant that just for the KMS implementation. 
> 
> > add additional APIs to interface to support io.Reader (not just []byte)
> 
> This 100%, too, but unfortunately it's not compatible with `crypto` libs and wouldn't have a practical upside unless the signer impl itself forgoes said libraries.

PGP libs do take an `io.Reader` but agree it won't help in most situations. 

> 
> What's the end goal?

I'd like to get to a single implementation within sigstore for signing and verification, which is what I think your original goal was here as well. There's a fair bit of duplicated code across sigstore where we're directly calling rsa/ecdsa instead of using the signature library, and having support for the `crypto.Signer` interface seems like it would help. 

There were two recent situations that triggered me to start working on this:

- While adding RFC3161 support in Rekor, @asraa ended up writing a fair bit of code to manually crafting messages to sign (instead of leveraging existing PKCS libraries which require passing a `crypto.Signer`). 
- If we wanted to embed SCTs in the Fulcio-generated CA certs, I need to sign the precert, and the `x509.CreateCertificate` call ([godoc](https://golang.org/src/crypto/x509/x509.go?s=69096:69213#L2130)) requires being passed a `crypto.Signer`. I can of course re-implement this, but would like to avoid that if at all possible.

+1 to `crypto.Signer` leaving a fair bit to be desired.

dekkagaijin:
What if we do both? Implement both `(signature, crypto).(Signer, Verifier)`? Should be straightforward since the existing interfaces are meant translate to their `crypto` cousins well, anyway.

bobcallaway:
> What if we do both? Implement both `(signature, crypto).(Signer, Verifier)`? Should be straightforward since the existing interfaces are meant translate to their `crypto` cousins well, anyway.

I think that makes sense. 

Let me mark this PR as a draft for now, and we can agree to the combined interface separately and then come back to this one.

dekkagaijin:
IMHO we'll want to split the high and low level interfaces. I'm not sure there's a good way to provide a complete interface that can translate down to `crypto.Signer` from these types

pkg/signature/verifier.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+type Verifier interface {
+	VerifySignature(signature []byte, message []byte, opts ...VerifierOption) error
+}
+
+type VerifyRequest struct {
```
dekkagaijin, line 35: I'm not happy that custom implementations of signers/verifiers would require type casting in order to allow for expanded options, but I also don't know of a better alternative /shrug

pkg/signature/ecdsa.go
```diff
@@ -1,17 +1,17 @@
 //
 // Copyright 2021 The Sigstore Authors.
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Licensed undee the Apache License, Version 2.0 (the "License");
```
dekkagaijin, line 4: looks like a search/replace error in the license test

pkg/signature/ecdsa.go
```diff
@@ -21,66 +21,293 @@ import (
...
+	if e.priv == nil {
+		return nil
+	}
+
+	return e.priv.Public()
+}
+
+// PublicWithContext returns the public key that can be used to verify signatures created by
+// this signer. As this value is held in memory, the context argument to this method is ignored.
+func (e ECDSASigner) PublicWithContext(_ context.Context) (crypto.PublicKey, error) {
```
dekkagaijin, line 146: we should probably genericize this to something like RPCOptions as well

pkg/signature/signer.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/pkg/errors"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+
+	// these ensure we have the implementations loaded
+	_ "golang.org/x/crypto/sha3"
+)
+
+type Signer interface {
+	SignMessage(message []byte, opts ...SignerOption) ([]byte, error)
+	crypto.Signer
```
dekkagaijin, line 41: I still firmly believe that our Signers shouldn't expose `crypto.Signer` directly, since there isn't a sane way to pass request context to implementations that require it

bobcallaway: I'm not sure I understand your objection on this point; in this PR, I leverage the fact that the third argument to `Sign(io.Reader, []byte, crypto.SignerOpts)` is an interface that we can implement and extend to pass the context into the call (done here in GCP KMS implementation: 
https://github.com/sigstore/sigstore/pull/69/files#diff-3a72d0bc4fa0dbf57003eab15443ab0c199be6dacb6c256fcd98609f60801335R159 )

If someone passes `signature.SignerOpts` which contained `WithContext(desiredCtx)`, then we can use it on the underlying external call to the KMS provider. If it's not provided by the caller of `Sign()` and we still need a context, then we could use the one set at the time the constructor was called.

pkg/signature/ecdsa.go
bobcallaway: oops, will fix that. good catch

pkg/signature/signer.go
dekkagaijin: The issue is `crypto.Signer` includes `PublicKey`, which doesn't have any args:
```
Public() PublicKey
```

dekkagaijin: IIRC one of the function examples that was the impetus for this change accepted a `crypto.Signer` but not any `crypto.SignerOpt`s, right? That would also break this scheme

dekkagaijin: Another option might be:

```
type ContextSigner struct {
  signature.Signer
  ctx context.Context
}

func (c ContextSigner) Public() crypto.PublicKey {
  pk, err := c.PublicWithContext(WithContext(c.ctx))
  if err != nil {
    panic(err)
  }
  return pk
}
```

bobcallaway:
@dekkagaijin After looking at what you put up in #70, and thinking it through a bit more, I've come to this (which I think achieves both of our goals):

- For the non-KMS signers (i.e. RSA/ECDSA/ED25519 operations are completed totally in-memory), we directly implement `crypto.Signer` but do not declare it as part of the `signature.Signer` interface. 
- For the KMS signers (that would need to have a context passed), I implemented a `CryptoSignerWrapper` like you suggested above. 
- I also re-vamped my options code to use the method you proposed in #70 since I do agree that its a bit more flexible in the long run than the `SignRequest`/`VerifyRequest` approach I started with.

I'd appreciate your comments on this latest update... thanks!

pkg/signature/sv_test.go
```diff
@@ -16,31 +16,143 @@
...
+
+	if err := v.VerifySignature(nil, bytes.NewReader(message)); err == nil {
+		t.Error("no error when passing nil as signature")
 	}
-	payload := []byte("test payload " + fmt.Sprint(rand.Int()))
-	sig, _, err := sv.Sign(ctx, payload)
-	if err != nil {
-		t.Fatalf("Sign() failed with error: %v", err)
+
+	if err := v.VerifySignature(bytes.NewReader([]byte("not the sig")), bytes.NewReader(message)); err == nil {
```
dekkagaijin, line 139:
```suggestion
	if err := v.VerifySignature(strings.NewReader("not the sig"), bytes.NewReader(message)); err == nil {
```

dekkagaijin (approved):
I like :)
Thanks for indulging me

lukehinds (approved):

PR #70, opened 2021-06-08
replace context with Options in Signer, Verifier, and PublicKeyProvider interfaces

dekkagaijin:
This allows `signature.Signers` to provide `crypto.Signer`s to common libraries.

### DIFF
--- a/cmd/sign.go
+++ b/cmd/sign.go
@@ -16,6 +16,7 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"crypto/x509"
 	"encoding/pem"
@@ -23,6 +24,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -31,6 +33,7 @@ import (
 	"github.com/sigstore/sigstore/pkg/httpclients"
 	"github.com/sigstore/sigstore/pkg/oauthflow"
 	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/option"
 	"github.com/sigstore/sigstore/pkg/tlog"
 	"github.com/sigstore/sigstore/pkg/utils"
 )
@@ -81,7 +84,7 @@ var signCmd = &cobra.Command{
 			return err
 		}
 
-		pub, err := signer.PublicKey(ctx)
+		pub, err := signer.PublicKey(option.WithContext(ctx))
 		if err != nil {
 			return err
 		}
@@ -92,7 +95,7 @@ var signCmd = &cobra.Command{
 			return err
 		}
 
-		proof, _, err := signer.Sign(ctx, []byte(idToken.Subject))
+		proof, err := signer.Sign(strings.NewReader(idToken.Subject), option.WithContext(ctx))
 		if err != nil {
 			return err
 		}
@@ -130,7 +133,7 @@ var signCmd = &cobra.Command{
 
 		fmt.Println("Received signing cerificate with serial number: ", cert.SerialNumber)
 
-		signature, signedVal, err := signer.Sign(ctx, payload)
+		signature, err := signer.Sign(bytes.NewReader(payload), option.WithContext(ctx))
 		if err != nil {
 			panic(fmt.Sprintf("Error occurred while during artifact signing: %s", err))
 		}
@@ -139,7 +142,6 @@ var signCmd = &cobra.Command{
 		fmt.Println("Sending entry to transparency log")
 		tlogEntry, err := tlog.UploadToRekor(
 			certPEM,
-			signedVal,
 			signature,
 			viper.GetString("rekor-server"),
 			payload,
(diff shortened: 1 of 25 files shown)


pkg/signature/iface.go
```diff
@@ -18,21 +18,30 @@ package signature
...
 type Signer interface {
 	PublicKeyProvider
 
-	// Sign takes a raw payload, potentially creating a derivative (e.g. hashed) payload, and returns the signature as well as the actual payload that was signed.
-	Sign(ctx context.Context, rawPayload []byte) (signature, signed []byte, err error)
+	// Sign takes a raw (not pre-hashed) message, and returns a `Verify`-able signature for the message.
+	Sign(ctx context.Context, rawMessage io.Reader) (signature []byte, err error)
+
+	// Returns a standard Golang `crypto.Signer` which uses the given context (if applicable) in addition to default SignerOpts and a MessagePreprocessor.
+	CryptoSigner(ctx context.Context) (crypto.Signer, crypto.SignerOpts, MessagePreprocessor, error)
```
dekkagaijin, line 44: @bobcallaway wdyt of this kind of interface?

dekkagaijin: /cc @asraa

bobcallaway: I think this will work; if all the caller has is the digest (instead of the entire message), its a bit awkward to have to get the `CryptoSigner()` to be able to sign; also we wouldn't be able to verify in that case but im not sure how frequent of a situation that is. 

I guess I'm still slightly in favor of the functional options approach (listed below since Jake and I iterated on it over Slack) - but I would be happy either way.

```
type Signer interface {
        SignMessage(message []byte, opts ...SignerOption) ([]byte, error)
        crypto.Signer
        // Public() crypto.PublicKey
        // Sign(rand io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error)
        PublicWithContext(context.Context) (crypto.PublicKey, error)
}

SignerOptions: WithContext(), WithDigest(), WithRand()
// WithContext(context.Context)    -> specifies the context for communication with external service
// WithRand(io.Reader)             -> specifies a source of entropy used during the signing process (defaults to crypto/rand.Reader if option is not specified)
// WithDigest([]byte, crypto.Hash) -> specifies the digest value of the message, along with the hash function used to compute it

type Verifier interface {
        VerifySignature(signature []byte, message []byte, opts ...VerifierOption) error
}

VerifierOptions: WithContext(), WithDigest()
// WithContext(context.Context)    -> specifies the context for communication with external service
// WithDigest([]byte, crypto.Hash) -> specifies the digest value of the message, along with the hash function used to compute it

// context could be set at the time of instantiation, or passed in via crypto.SignerOpts like this:
type GCPContextSignerOpts struct {
        Hash    crypto.Hash
        Context context.Context
}
```

dekkagaijin: > if all the caller has is the digest (instead of the entire message), its a bit awkward to have to get the CryptoSigner() to be able to sign; also we wouldn't be able to verify in that case but im not sure how frequent of a situation that is.

IMHO we want to abstract those implementation details away for the high-level interfaces. There'll definitely be use-cases where those details are relevant complexity, but I don't think we should make the complexity mandatory in order to support them

lukehinds: I tend to agree with Bob's approach here, as this allows a simple interface with no mandatory args to pass (with no need to `nil` out values), so it could be no args to just sign a message, or if required a specific context (as one example), is required, we have the flexibility to pass in exactly what we need.

lukehinds: @dlorenc  / @asraa any thoughts?

dlorenc: Functional arguments work for me, I'm not picky though.

asraa: Hm I think the outline Bob has above will make it very easy to use CryptoSigner. + 1

I might be missing something, why even have `SignMessage`?

bobcallaway: `SignMessage` is there as a counterpart to `VerifyMessage`, so that given a payload and a `SignerVerifier sv`, you should be able to perform the combination of:

`sv.VerifyMessage(sv.SignMessage(message),message)`

(this exact syntax isn't possible due to error handling, but hopefully the intent is clear).

bobcallaway: https://github.com/sigstore/sigstore/pull/69 has been updated to implement the functional options approach discussed above... comments welcome!

dekkagaijin: I'm happy enough with everything in #69 other than making our Signers also declare crypto.Signer. That's not to say that the specific implementations can't or shouldn't implement `crypto.Signer` directly, but IMHO there should be a sane way to pass request context to an implementation that requires it. PTAL at my proposal for dealing with that: https://github.com/sigstore/sigstore/pull/70/files#diff-d408de98cc9d5ef6f06ab23cdeab573b79f589766bcdf05ebf207628f8ac63f6

```
// Returns a standard Golang `crypto.Signer` which uses the given RPC options (if applicable) in addition to default SignerOpts and a MessagePreprocessor.
CryptoSigner(rpcOpts ...RPCOption) (crypto.Signer, error)
```

dekkagaijin:
reconciled in #69

PR #74, opened 2021-07-01
AWS KMS Support

codysoyland:
This adds AWS as a supported KMS backend.

### DIFF
--- a/pkg/signature/kms/aws/client.go
+++ b/pkg/signature/kms/aws/client.go
@@ -1,0 +1,319 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package aws
+
+import (
+	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rsa"
+	"crypto/x509"
+	"fmt"
+	"io"
+	"regexp"
+	"time"
+
+	"github.com/ReneKroon/ttlcache/v2"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/pkg/errors"
+	"github.com/sigstore/sigstore/pkg/signature"
+)
+
+const (
+	CacheKey        = "signer"
+	ReferenceScheme = "awskms://"
+)
+
+type awsClient struct {
+	client        *kms.KMS
+	keyResourceID string
+	endpoint      string
+	keyID         string
+	alias         string
+	keyCache      *ttlcache.Cache
+}
+
+var (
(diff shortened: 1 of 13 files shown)


pkg/signature/kms/aws/signer.go
```diff
@@ -0,0 +1,236 @@
...
+// This function recognizes the following Options listed in order of preference:
+//
+// - WithDigest()
+//
+// - WithCryptoSignerOpts()
+//
+// - WithContext()
+//
+// All other options are ignored if specified.
+func (a *SignerVerifier) VerifySignatureRemotely(sig, message io.Reader, opts ...signature.VerifyOption) (err error) {
```
bobcallaway, line 158: I like that you're giving the choice here, but I'm wondering whether we should add a new option `WithRemoteVerification(bool)` that flips between local or remote verification instead of having a different method signature. I could be convinced either way, but just adding the thought for discussion.

bobcallaway (commented):
I recently updated the KMS interface to add two new methods (that are largely useful during the CreateKey call) https://github.com/sigstore/sigstore/blob/5b6fffda16e4b2e0746c05729e05b59bc79baded/pkg/signature/kms/kms.go#L73-L74
Could you please add those as well - otherwise LGTM!

pkg/signature/kms/aws/client.go
```diff
@@ -0,0 +1,319 @@
...
+	return
+}
+
+func (a *awsClient) setupClient() (err error) {
+	var sess *session.Session
+	if a.endpoint == "" {
+		sess, err = session.NewSession()
+	} else {
+		sess, err = session.NewSession(&aws.Config{
+			Endpoint: aws.String("http://" + a.endpoint),
```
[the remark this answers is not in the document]
codysoyland: I was wondering about that... This is mostly in place for testing, but if there is a use-case for custom endpoints (KMS-compatible providers), then https would surely be preferred, but we might need to add some way to disable TLS verification in testing, maybe with an env var? I can look into that.

pkg/signature/kms/aws/signer.go
codysoyland: Good call, I think that would be cleaner.

pkg/signature/kms/aws/signer.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return a.client.verifyRemotely(ctx, sigBytes, digest)
+}
+
+// CreateKey attempts to create a new key in Vault with the specified algorithm.
+func (a *SignerVerifier) CreateKey(ctx context.Context, algorithm string) (crypto.PublicKey, error) {
+	return a.client.createKey(ctx, algorithm)
+}
+
+type cryptoSignerWrapper struct {
```
codysoyland, line 203: This is duplicated from the other two KMS backends... should we factor this out into a different package?

bobcallaway: probably, but we could do that in a separate PR

pkg/signature/kms/aws/client.go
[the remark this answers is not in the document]
codysoyland: Latest change switches to `https` for custom endpoints and adds `AWS_TLS_INSECURE_SKIP_VERIFY` env var for testing.

pkg/signature/kms/aws/signer.go
codysoyland: Added that option in latest commit.

bobcallaway:
LGTM - will approve once the e2e tests pass

codysoyland:
Thanks @bobcallaway for your quick reviews! My last patch should fix the e2e test failures.

codysoyland:
After that last failure, I set up GHA on my fork and realized it wasn't running the e2e-test.sh script, so I updated the GHA config with the localstack settings and got it passing there. It should be good now! 🤞

bobcallaway (approved):

PR #75, opened 2021-07-02
Add PublicKeyProvider interface

bobcallaway:
Given the requirements of cosign, I split out the PublicKey() method into its own interface such that it can be implemented by both signers and verifiers (since both have a public key).

Also small fixes including:
 - ed25519 passes by value rather than pointer for public/private keys
 - adds SupportedAlgorithms() and DefaultAlgorithm() method for KMS SignerVerifiers (ideal for use in CreateKey() method)

Signed-off-by: Bob Callaway <bob.callaway@gmail.com>

### DIFF
--- a/pkg/signature/verifier.go
+++ b/pkg/signature/verifier.go
@@ -29,6 +29,7 @@ import (
 )
 
 type Verifier interface {
+	PublicKeyProvider
 	VerifySignature(signature, message io.Reader, opts ...VerifyOption) error
 }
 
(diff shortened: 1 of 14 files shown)


bobcallaway:
@dekkagaijin could you please take a look? thanks!

pkg/signature/verifier.go
```diff
@@ -43,8 +44,8 @@ func LoadVerifier(publicKey crypto.PublicKey, hashFunc crypto.Hash) (Verifier, e
 		return LoadRSAPKCS1v15Verifier(pk, hashFunc)
 	case *ecdsa.PublicKey:
 		return LoadECDSAVerifier(pk, hashFunc)
-	case *ed25519.PublicKey:
-		return LoadED25519Verifier(*pk)
+	case ed25519.PublicKey:
```
lukehinds, line 47: I guess you meant to remove the pointer here (line 47)?

lukehinds (approved):
LGTM, just a small note.

pkg/signature/verifier.go
bobcallaway: yes, this was intentional. `rsa` and `ecdsa` represent their public key in structs, but `ed25519` just has a `[]byte` representation so they're handled differently.

PR #76, opened 2021-07-04
Add initial Azure KMS support

cpanato:
Add initial Azure KMS support, Just saw now that there are some refactors to do. I will apply those. 

Opening as a draft for gathering any feedback


pkg/kms/azure/azure.go
```diff
@@ -0,0 +1,240 @@
...
+	return
+}
+
+func NewAzureKMS(ctx context.Context, keyResourceID string) (*KMS, error) {
+	vaultURL, vaultName, keyName, err := parseReference(keyResourceID)
+	if err != nil {
+		return nil, err
+	}
+
+	azureTentatID := os.Getenv("AZURE_TENANT_ID")
```
dlorenc, line 78: typo, tentat->tenant

pkg/kms/azure/azure.go
```diff
@@ -0,0 +1,240 @@
...
+	if azureTentatID == "" {
+		return nil, errors.New("AZURE_TENANT_ID is not set")
+	}
+
+	azureClientID := os.Getenv("AZURE_CLIENT_ID")
```
dlorenc, line 83: I don't know much about Azure - are these typically set as environment variables? Would it make sense for them to be passed as flags instead?

dlorenc (commented):
This looks great! One question about the env vars.

pkg/kms/azure/azure.go
cpanato: I was following a similar approach that I saw for vault. I will review that when apply the rebate and refactor to the new way

Thanks for you review and feedback

dlorenc: Thanks! I asked the same question on the vault one, and I guess that environment variable is the standard, canonical method of specifying the address, so it made sense to keep.

lukehinds: I think viper.GetString works with both flags and env vars?

containerpope: @dlorenc @cpanato from what I remember, in Azure the default is by environment variables, alternatively a file could be used or cli device login. https://github.com/Azure/azure-sdk-for-go#more-authentication-details

The env based login approach that is currently implemented is usually used for application based login with either an Azure Application (client secret/cert) or Azure Managed Service Identity (works only on azure based vms and services, where this is available) is used for the login.

I think for the cosign cli it would be valuable to implement also the **auth.NewAuthorizerFromCli()**, which then uses an installed azure cli to execute a device login, where a browser is openend and the user can do his single sign on. With this method not only technical users, but also regular cli users could run the signing. 
Maybe this could be toggled via cli flag.

cpanato: I had that in my mind as well, we also use the env var is in CAPZ (cluster API for azure)

Is that ok to keep as-is for this iteration? @dlorenc

containerpope: I think also that a specific sign in by user might be a use case, that is not that relevant when only a CI is doing the signing. Maybe something to consider for the future :)

cpanato:
integrated with cosign

[asciicast](https://asciinema.org/a/CaLovXymaiWrU8ErtqqbSpZuH)

cpanato:
@dlorenc this is an initial implementation, I'm sure there are lots of things to improve. But I think we can do any followups and also maybe other persons more expert than I can help here as well :)

the video above shows the commands with cosign to generate the key, sign, and verify! :)

let me know what you think

dlorenc (approved):

dlorenc:
Awesome!

PR #78, opened 2021-07-06
Fix for Error: error during PEM decoding

lukehinds:
Break condition was trying to match an empty array with nil

This resulted in no break condition and the error working down
to `error during PEM decoding`

Instead match on an empty length


cpanato (approved):

bobcallaway (approved):
LGTM

PR #101, opened 2021-07-22
ENCRYPTED COSIGN PRIVATE KEY -> ENCRYPTED SIGSTORE PRIVATE KEY

dekkagaijin:

### DIFF
--- a/pkg/cryptoutils/privatekey.go
+++ b/pkg/cryptoutils/privatekey.go
@@ -31,6 +31,7 @@ import (
 
 const (
 	PrivateKeyPEMType                  PEMType = "PRIVATE KEY"
+	encryptedCosignPrivateKeyPEMType   PEMType = "ENCRYPTED COSIGN PRIVATE KEY"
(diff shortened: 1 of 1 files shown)


pkg/cryptoutils/privatekey.go
```diff
@@ -102,7 +102,7 @@ func UnmarshalPEMToPrivateKey(pemBytes []byte, pf PassFunc) (crypto.PrivateKey,
 	switch derBlock.Type {
 	case string(PrivateKeyPEMType):
 		return x509.ParsePKCS8PrivateKey(derBlock.Bytes)
-	case string(EncryptedCosignPrivateKeyPEMType):
+	case string(EncryptedSigstorePrivateKeyPEMType):
```
dlorenc, line 105: Can we support both here? It's kinda ugly but it it decodes, it decodes.

dekkagaijin: sgtm

dlorenc (approved):

PR #102, opened 2021-07-22
Transit backend path is hardcoded for some operations of the KMS Vault client

LeSuisse:
The environment variable TRANSIT_SECRET_ENGINE_PATH is not always used.

Some operations are hardcoded to "transit/". Users with a transit backend mounted with a non default path might not be able to use it.


bobcallaway (approved):
LGTM

PR #104, opened 2021-07-25
leverage Vault token helpers approach while obtaining Vault token

developer-guy:
There is a concept called [Token Helpers](https://www.vaultproject.io/docs/commands/token-helper) in Vault. By default, the Vault CLI will store the generated token on disk in the `~/.vault-token` file. So, I realized that even though we could not set the `VAULT_TOKEN` env variable, we can still obtain the Vault token from the file with the help of this feature. So, this PR will add that support to the Vault client initialization process.

Btw, this is just an idea, so please feel free to comment it out.

Screen Shot 2021-07-25 at 11 58 14

Screen Shot 2021-07-25 at 11 58 29

### DIFF
--- a/pkg/signature/kms/hashivault/client.go
+++ b/pkg/signature/kms/hashivault/client.go
@@ -21,13 +21,16 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"log"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
 
 	"github.com/ReneKroon/ttlcache/v2"
 	vault "github.com/hashicorp/vault/api"
+	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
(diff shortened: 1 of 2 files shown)


pkg/signature/kms/hashivault/client.go
```diff
@@ -84,7 +87,18 @@ func newHashivaultClient(keyResourceID string) (*hashivaultClient, error) {
...
-		return nil, errors.New("VAULT_TOKEN is not set")
+		log.Printf("VAULT_TOKEN is not set, trying to read token from file at path ~/.vault-token")
+		homeDir, err := homedir.Dir()
+		if err != nil {
+			return nil, err
+		}
+
+		tokenFromFile, err := os.ReadFile(filepath.Join(homeDir, ".vault-token"))
+		if err != nil {
+			return nil, err
```
cpanato, line 98: maybe wrap the error here to give more context

developer-guy: thank you so much for this feedback, I updated the code according to your feedback. 👍

cpanato (approved):
looks great to me :)

pkg/signature/kms/hashivault/client.go
```diff
@@ -94,6 +108,8 @@ func newHashivaultClient(keyResourceID string) (*hashivaultClient, error) {
 		return nil, errors.Wrap(err, "new vault client")
 	}
 
+	client.SetToken(token)
```
bobcallaway, line 111: just curious - why are we doing this now VS not before?

developer-guy: hello @bobcallaway, by default, the Vault client is using the `VAULT_TOKEN` environment variable implicitly, so we just check the existence of the environment variable instead of using it explicitly, but now we are reading this from a file, so we have to set it explicitly.

bobcallaway: makes sense, thanks @developer-guy!

PR #106, opened 2021-07-26
default to P-256 curve again

dekkagaijin:
After discussion regarding also. defaulting to P-384 in Cosign, we decided to stick with P-256 due to broad popularity, support, and performance. https://malware.news/t/everyone-loves-curves-but-which-elliptic-curve-is-the-most-popular/17657

Not _too_ opinionated, but would prefer to err on the side of consistency unless there's a good reason why the default was change to P-384 in the first place

### DIFF
--- a/pkg/signature/ecdsa.go
+++ b/pkg/signature/ecdsa.go
@@ -196,9 +196,9 @@ func LoadECDSASignerVerifier(priv *ecdsa.PrivateKey, hf crypto.Hash) (*ECDSASign
 
 // NewDefaultECDSASignerVerifier creates a combined signer and verifier using ECDSA.
 //
-// This creates a new ECDSA key using the P-384 curve and uses the SHA256 hashing algorithm.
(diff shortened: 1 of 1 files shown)


bobcallaway (approved):
LGTM

dekkagaijin:
A more recent report seems to suggest the popularity of 256-bit curves is still vastly higher: https://www.f5.com/content/dam/f5-labs-v2/article/pdfs/F5Labs-2019-TLS-Telemetry-Report-Summary.pdf

lukehinds:
I don't have a strong view on the curve type,  but thought we might have aglo agility for this and tools can pass i what they want?

lukehinds:
oh, already merged

bobcallaway:
we do have agility for this - there is a method: `NewECDSASignerVerifier` where you can specify a non-default curve

dekkagaijin:
@lukehinds agreed on not being too opinionated here, but it seems like P-256 (and X25519) are nearly 100x as popular as alternative curves. As Bob mentioned, you can choose a non-default curve with the non-default constructor

PR #109, opened 2021-08-25
go: upgrade golang to 1.17

cpanato:
go: upgrade golang to 1.17


lukehinds (changes requested):
same comment as per fulcio, would rather not change unless we really have specific need.

cpanato:
closing for now

PR #118, opened 2021-10-05
Modularise CI

lukehinds:
Make tests more easily readable by modularise into specific
roles (inline with rekor/ cosign ..)

It also adds CodeQL checks

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -26,6 +26,7 @@ env:
 
 jobs:
   e2e_test:
+    name: e2e integration tests
     runs-on: ubuntu-latest
     services:
       vault:
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -22,7 +22,7 @@ on:
     branches: [ main ]
 
 env:
-  GO_VERSION: 1.16.6
+  GO_VERSION: 1.16.8
 
 jobs:
   e2e_test:
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -13,10 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Go
-
-permissions:
-  contents: read
+name: e2e-test
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


lukehinds:
note `build Expected — Waiting for status to be reported` is from branch protection (I will tweak these if / when this merges)

.github/workflows/e2e_test.yml
```diff
@@ -13,10 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Go
-
-permissions:
-  contents: read
+name: test
```
cpanato, line 16: would call this e2e-test instead of test, we already have a test job below

.github/workflows/test.yml
```diff
@@ -0,0 +1,56 @@
...
+  contents: read
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  GO_VERSION: 1.16.6
```
cpanato, line 28:
```suggestion
  GO_VERSION: 1.16.8
```

.github/workflows/e2e_test.yml
```diff
@@ -28,42 +25,9 @@ env:
   GO_VERSION: 1.16.6
```
cpanato, line 25:
```suggestion
  GO_VERSION: 1.16.8
```

cpanato:
@lukehinds Also need to check the status check in the settings, I don't have this permission to update it but need to remove the `build` and add the others as required

.github/workflows/e2e_test.yml
lukehinds: This is by design as it groups tests, if we don't have this it shows up as "e2e-test e2e tests"..

image

cpanato: ahh, i see what you mean! sounds great then!

lukehinds:
> @lukehinds Also need to check the status check in the settings, I don't have this permission to update it but need to remove the `build` and add the others as required
> 
> except for the `codeQL` I don't think that needs to be required to merge a PR

yep https://github.com/sigstore/sigstore/pull/118#issuecomment-934254238

cpanato (approved):
thanks!

PR #123, opened 2021-10-19
Idp specific default flows

houdini91:
Suggestion:
Adding interactive flows for each specific  identity provider, allowing users to skip the main idp selection page.
I think one less click can improve the UX a bit.
Further UX improvement is gained when browser uses a default idp account the user does may not need to interactively intervene at all.
Such psodo-auto flow may also be valuable in automation uses cases, for example a git hook signing SLSA  provenance.

Hope this helps .
mikey strauss

### DIFF
--- a/pkg/oauthflow/flow.go
+++ b/pkg/oauthflow/flow.go
@@ -37,9 +37,10 @@ const (
 </html>
 `
 
-	GithubAuthSubURL    = "https://github.com/login/oauth"
-	GoogleAuthSubURL    = "https://accounts.google.com"
-	MicrosoftAuthSubURL = "https://login.microsoftonline.com"
+	// Default connector ids used by `oauth2.sigstore.dev` for specfic Idps.
+	PublicInstanceGithubAuthSubURL    = "https://github.com/login/oauth"
+	PublicInstanceGoogleAuthSubURL    = "https://accounts.google.com"
+	PublicInstanceMicrosoftAuthSubURL = "https://login.microsoftonline.com"
 )
 
 type TokenGetter interface {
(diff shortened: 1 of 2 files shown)


houdini91:
@bobcallaway 
Fixed your comments,
flow.go examples work and may be useful for library users (such as cosign).
I would leave them if it was up to me, but i thought you may not want them.

houdini91:
Turns out #https://github.com/sigstore/fulcio/pull/204 broke the connector_id argument.
I am closing this PR.

bobcallaway:
the connector IDs did change, but I think this patch is still valid if we update them to match...

houdini91:
@bobcallaway 
If you think it's still possible I be happy to continue.
Can you point me on the right direction?

houdini91:
@bobcallaway 
i think understand correct me if i am wrong.


I can use the new connector_ids values to choose between idps.
I tested the new Ids and they seem to work.
```
https://github.com/login/oauth
https://accounts.google.com
https://login.microsoftonline.com
```
After the change  do the Default<privider>IDTokenGetter make more sense as part of the code base?

bobcallaway:
> After the change do the Default(privider)IDTokenGetter make more sense as part of the code base?

I think its fine if you want to put them in.

pkg/oauthflow/flow.go
```diff
@@ -36,6 +37,11 @@ const htmlPage = `<html>
 </html>
 `
```
bobcallaway, line 39: Perhaps we should add some comments to these (as well as updating the names) to denote that they are only meant to be used with `oauth2.sigstore.dev` ; otherwise this LGTM

houdini91: I updated a bit of names and comments
Did you mean renaming the Default[provider]IDTokenGetter as well?
Is there any specific naming prefix you would rather?
PublicInstance, SigstoreDev, .. ?

bobcallaway: I'm fine with `PublicInstance` as you propose here. I think with renaming the IDTokenGetters we'd be ready to merge this. Thanks for the persistence on getting this to the finish line :)

houdini91: No problem, its my pleasure thanks for the review.
Its my first contribution slowly i hope to have help out.
If you like you can take a look at other PR's i have open on cosign and go-securesystemslib.

bobcallaway (approved):

PR #124, opened 2021-10-29
Make SimpleContainerImage struct accesible for tekton chains

priyawadhwa:
Right now Chains is using its own custom `SimpleSigning` struct for container images, which is marshaling JSON differently from how cosign does it. I want to copy cosign and use this library in chains as well, exposing this function should make it easier to implement.

Signed-off-by: Priya Wadhwa <priyawadhwa@google.com>

-->
```release-note
NONE
```


lukehinds (approved):
LGTM

PR #125, opened 2021-10-29
(fix): Fix vault integration to work with rotated keys

rjbrown57:
Signed-off-by: Russell Brown <rjbrown57@gmail.com>

Fixes https://github.com/sigstore/cosign/issues/890

* Use regex instead of trimPrefix to fix signing
* Introduce environment variable VAULT_KEY_PREFIX to allow setting of which key version to verify with. If not specified will default to "vault:v1:"

### DIFF
--- a/pkg/signature/kms/hashivault/client.go
+++ b/pkg/signature/kms/hashivault/client.go
@@ -25,7 +25,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strings"
 	"time"
 
 	"github.com/ReneKroon/ttlcache/v2"
@@ -45,6 +44,7 @@ type hashivaultClient struct {
 var (
 	errReference   = errors.New("kms specification should be in the format hashivault://<key>")
 	referenceRegex = regexp.MustCompile(`^hashivault://(?P<path>\w(([\w-.]+)?\w)?)$`)
+	prefixRegex    = regexp.MustCompile("vault:v[0-9]+:")
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/signature/kms/hashivault/client.go
```diff
@@ -48,7 +47,7 @@ var (
 )
 
 const (
-	vaultV1DataPrefix = "vault:v1:"
+	vaultV1DataRegExPrefix = "vault:v[0-9]+:"
```
bobcallaway, line 50: Nit: use `regexp.MustCompile` as above so we don’t recompile the regex every time we call `vaultDecode`

pkg/signature/kms/hashivault/client.go
```diff
@@ -200,9 +199,14 @@ func (h hashivaultClient) verify(sig, digest []byte, alg crypto.Hash) error {
 	client := h.client.Logical()
 	encodedSig := base64.StdEncoding.EncodeToString(sig)
 
+	vaultDataPrefix := os.Getenv("VAULT_KEY_PREFIX")
+	if vaultDataPrefix == "" {
+		vaultDataPrefix = "vault:v1:"
```
bobcallaway, line 204: Define this as a `const` above?

bobcallaway (commented):
LGTM minus two small nits

pkg/signature/kms/hashivault/client.go
rjbrown57: Adjusted as requested. Thanks for the good feedback!

bobcallaway (approved):

PR #126, opened 2021-11-06
Fix the azure KMS provider

dlorenc:
There were two issues: double hashing and the wrong signature formats.

We already hash the contents in SignMessage using a configurable
hash function, so we don't need this one here.

The other issue was caused by azure returning and expecting a raw byte sequence
of R||S concatenated, rather than an ASN.1 sequence. Big thanks to mwielgoszewski
for providing the tips and POC.

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/signature/kms/azure/signer.go
+++ b/pkg/signature/kms/azure/signer.go
@@ -20,8 +20,11 @@ import (
 	"context"
 	"crypto"
 	"io"
+	"math/big"
 
 	"github.com/pkg/errors"
+	"golang.org/x/crypto/cryptobyte"
+	"golang.org/x/crypto/cryptobyte/asn1"
 
 	"github.com/sigstore/sigstore/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/signature/options"
@@ -104,7 +107,25 @@ func (a *SignerVerifier) SignMessage(message io.Reader, opts ...signature.SignOp
 		return nil, err
 	}
 
-	return a.client.sign(ctx, digest)
+	rawSig, err := a.client.sign(ctx, digest)
+	if err != nil {
+		return nil, err
+	}
+
+	l := len(rawSig)
+	r, s := &big.Int{}, &big.Int{}
+	r.SetBytes(rawSig[0 : l/2])
+	s.SetBytes(rawSig[l/2:])
+
+	// Convert the concantenated r||s byte string to an ASN.1 sequence
+	// This logic is borrowed from https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/crypto/ecdsa/ecdsa.go;l=121
+	var b cryptobyte.Builder
+	b.AddASN1(asn1.SEQUENCE, func(b *cryptobyte.Builder) {
+		b.AddASN1BigInt(r)
+		b.AddASN1BigInt(s)
+	})
+
+	return b.Bytes()
 }
 
 // VerifySignature verifies the signature for the given message. Unless provided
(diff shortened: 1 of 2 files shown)


pkg/signature/kms/azure/signer.go
```diff
@@ -104,7 +107,25 @@ func (a *SignerVerifier) SignMessage(message io.Reader, opts ...signature.SignOp
 		return nil, err
 	}
 
-	return a.client.sign(ctx, digest)
+	rawSig, err := a.client.sign(ctx, digest)
+	if err != nil {
+		return nil, err
+	}
+
+	l := len(rawSig)
```
dlorenc, line 115: This part feels gross but I couldn't think of another way to split the byte sequence.

@mwielgoszewski - any better ideas here?

cpanato: will test it

cpanato: did the test and now it is working, this is awesome Dan

```shell
$ ./cosign sign --key azurekms://cpanato-test-vault.vault.azure.net/test-123 -a TEST=123 ctadeu/test-sign:3

$ ./cosign verify --key azurekms://cpanato-test-vault.vault.azure.net/test-123 ctadeu/test-sign:3

Verification for index.docker.io/ctadeu/test-sign:3 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.

[{"critical":{"identity":{"docker-reference":"index.docker.io/ctadeu/test-sign"},"image":{"docker-manifest-digest":"sha256:0d1654f14007b1a6b3dda021323bcc8841cf580dab340f9b2c17ad71b184e866"},"type":"cosign container image signature"},"optional":{"TEST":"123"}}]

$ ./cosign verify --key azure.pub ctadeu/test-sign:3

Verification for index.docker.io/ctadeu/test-sign:3 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - The signatures were verified against the specified public key
  - Any certificates were verified against the Fulcio roots.

[{"critical":{"identity":{"docker-reference":"index.docker.io/ctadeu/test-sign"},"image":{"docker-manifest-digest":"sha256:0d1654f14007b1a6b3dda021323bcc8841cf580dab340f9b2c17ad71b184e866"},"type":"cosign container image signature"},"optional":{"TEST":"123"}}]
```

the pub key is

```
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEQ+oLBcTtoWxmVQJmeLKJxEwBxpqr
mqeZ+k1kzgoOOJdpjA+FBSxQPj82ov7eezYqOLfhKtGHw7SWP9JZ6V+FSQ==
-----END PUBLIC KEY-----
```

mwielgoszewski: > This part feels gross but I couldn't think of another way to split the byte sequence.
> 
> @mwielgoszewski - any better ideas here?

Either divide in half like you have, or use a constant based on the key/curve signature size (if you plan to support keys other than P-256). FWIW, the Azure SDK just [floor divides the raw signature length in half](https://github.com/Azure/azure-sdk-for-python/blob/e7ce98f5c68e45b20b7653c3e5d857355cbf06f8/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/_internal.py#L83).

bobcallaway (approved):
LGTM

PR #141, opened 2021-11-14
Move the ssh signing/verification utilities to sigstore from rekor.

dlorenc:
SSH signing has started to gain momentum with support in git! We have an implementation
in rekor that works, but it probably doesn't belong there. Let's move it here and realign
it with our sigstore signing APIs.

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/signature/ssh/README.md
+++ b/pkg/signature/ssh/README.md
@@ -1,0 +1,118 @@
+# SSH File Signatures
+
+SSH keys can be used to sign files!
+Unfortunately this is a pretty recent change to the openssh tooling, so it is not
+supported by golang.org/x/crypto/ssh yet.
+
+This document explains how it works at a high level.
+
+## Keys
+
+SSH keys are usually split into public and private files, named `id_rsa.pub` and
+`id_rsa`, respectively.
+These files are encoded and formatted a little differently than other signing keys.
+
+### Public Keys
+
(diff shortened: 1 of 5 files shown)


bobcallaway:
I like this (and am supportive of moving over the rest of `rekor/pki` where it makes sense). I think we would still need a way to generate a canonicalized version of a public key and signature to persist in the log (which is what is missing in `sigstore/sigstore` right now.

dlorenc:
> I like this (and am supportive of moving over the rest of `rekor/pki` where it makes sense). I think we would still need a way to generate a canonicalized version of a public key and signature to persist in the log (which is what is missing in `sigstore/sigstore` right now.

Yeah I don't think we need to move allllll of it over. This one was particularly ripe for a move though because we don't even use the Sign function other than for tests of Verify.

dekkagaijin (dismissed):
LGTM FWIW

lukehinds:
also lgtm, do we need to coordinate a removal PR from rekor?

lukehinds (approved):

dlorenc:
Thanks for the reminders - cleaned up the last TODOs.

dlorenc:
PTAL, should be good now!

lukehinds (approved):

Issue #172, opened 2021-12-01
Failed to verify signature in DSSE


naveensrinivasan:
**Description**

Part of fuzz testing DSSE verify is failing for this data`"\xb3"`

```
panic: failed to verify signature

goroutine 1 [running]:
github.com/sigstore/sigstore/pkg/signature/dsse.Fuzz({0x7fa1488e6000, 0x1, 0x0})
	/home/sammy/go/src/github.com/naveensrinivasan/sigstore/pkg/signature/dsse/fuzz.go:56 +0x757
go-fuzz-dep.Main({0xc0000dff68, 0x1, 0x575300})
	go-fuzz-dep/main.go:36 +0x15b
main.main()
	github.com/sigstore/sigstore/pkg/signature/dsse/go.fuzz.main/main.go:15 +0x3b
exit status 2%
```

naveensrinivasan:
@dlorenc FYI..

dlorenc:
Cc @bobcallaway I think you were trying to reproduce this right?

bobcallaway:
The root issue here is that with a `payloadType` of `\xbe` (note: not a valid UTF-8 string), different values are used before and after marshaling into the JSON envelope format (see [String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune.](https://pkg.go.dev/encoding/json#Marshal). 

The change in the `payloadType` value caused by golang's `json.Marshal` and `json.Unmarshal` is the bug here and probably should be filed upstream.

PR #173, opened 2021-12-01
Fuzz testing DSSE

naveensrinivasan:


pkg/signature/dsse/dsse_test.go
```diff
@@ -41,8 +41,8 @@ func TestRoundTrip(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	data := "sometestdata"
-	payloadType := "foo"
+	data := "\xb3"
```
naveensrinivasan, line 44: This is the failing case.

naveensrinivasan:
Related to https://github.com/sigstore/sigstore/issues/172

pkg/signature/dsse/dsse_test.go
naveensrinivasan: @dlorenc FYI...looks like you have implemented the DSSE. It is failing for this data

dlorenc: Do you have the failure handy? It might actually be in the upstream library (which I also implemented) over here: https://github.com/secure-systems-lab/dsse

naveensrinivasan: If the payloadType is set to `\xb3` it fails.

bobcallaway (approved):

PR #228, opened 2022-01-14
KMS: Change how the Azure authentication method is handled

### LINKED ISSUE #223: Adding more authentication options to Azure KMS
**Description**

Hi!

Great job you have been doing here! 😊

When looking at cosign I noticed that the Azure KMS integration only supports a service principal through the environment - but there are lots of other methods available from the Azure SDK.

Is there a specific reason to only support one of them or would you be open to a PR that adds support for MSI (Managed Service Identity) and Azure CLI (if you are authenticated using the cli, use the credentials from there).

I would be willing to try and add some additional support if you want! 👍

simongottschlag:
#### Summary
This patch removes the requirement of having the environment variables `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` set to use the Azure KMS (KeyVault).

By removing the requirement, we enable usage of MSI (Managed Service Identity) through the `NewAuthorizerFromEnvironment()` and we at the same time add support for the `NewAuthorizerFromCLI()`.

By splitting the function of calculating what method to use to it's own function , `getAuthenticationMethod()`, it's possible to test the logic separately.

We also introduce the new environment variable `AZURE_AUTH_METHOD` which if set to `environment` will use `FromEnvironment()` (may be useful for the MSI case) and if set to `cli` will use `FromCLI()`.

If nothing is defined, `FromEnvironment()` will be tested first and then `FromCLI()`. :^)

#### Ticket Link
Fixes #223 

#### Release Note
```release-note
Add support for more Azure KMS authentication methods.
```

### DIFF
--- a/pkg/signature/kms/azure/client.go
+++ b/pkg/signature/kms/azure/client.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 The Sigstore Authors.
+// Copyright 2022 The Sigstore Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -33,6 +33,7 @@ import (
 
 	kvauth "github.com/Azure/azure-sdk-for-go/services/keyvault/auth"
 	"github.com/Azure/azure-sdk-for-go/services/keyvault/v7.1/keyvault"
+	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
@@ -82,21 +83,6 @@ func newAzureKMS(ctx context.Context, keyResourceID string) (*azureVaultClient, 
 		return nil, err
 	}
 
-	azureTenantID := os.Getenv("AZURE_TENANT_ID")
-	if azureTenantID == "" {
-		return nil, errors.New("AZURE_TENANT_ID is not set")
-	}
-
-	azureClientID := os.Getenv("AZURE_CLIENT_ID")
-	if azureClientID == "" {
-		return nil, errors.New("AZURE_CLIENT_ID is not set")
-	}
-
-	azureClientSecret := os.Getenv("AZURE_CLIENT_SECRET")
-	if azureClientSecret == "" {
-		return nil, errors.New("AZURE_CLIENT_SECRET is not set")
-	}
-
 	client, err := getKeysClient()
 	if err != nil {
 		return nil, errors.Wrap(err, "new azure kms client")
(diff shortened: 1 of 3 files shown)


cpanato:
the current approach still works?

pkg/signature/kms/azure/client_test.go
```diff
@@ -0,0 +1,90 @@
+package azure
```
cpanato, line 1: please add the license header file, check the other go files for that

simongottschlag:
@cpanato yes, but please verify for yourself as well since I'm quite new to the project and haven't used it extensively yet. But the main point of the excercise of building the tests was to make sure we're getting the correct auth method based on the previous configuration - meaning if you already have `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` configured you will continue to use `FromEnvironment()`. But if you don't define them, it will start with `FromEnvironment()` and then test `FromCLI()` if the first one returned an error. I've also added a mechanism (`AZURE_AUTH_METHOD`) to short circuit the logic if defined using either `environment` or `cli`.

pkg/signature/kms/azure/client_test.go
simongottschlag: Fixed!

simongottschlag:
@cpanato I've built cosign locally with it pointing `sigstore` to my local branch. Noted that I had missed to remove the requirements and that's fixed now. Seems to work fine with both client credentials and CLI. :+1:

cpanato:
i will do some tests during the weekend! thanks for doing this

Are you willing to add some documentation about that here https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/azure/README.md ?

pkg/signature/kms/azure/client.go
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 The Sigstore Authors.
+// Copyright 2021-2022 The Sigstore Authors.
```
cpanato, line 2: just 2022 is ok

simongottschlag: Fixed!

simongottschlag:
@cpanato I've updated the readme with some additional information. Took the liberty to add some links and example commands while I was at it.

dlorenc:
Thanks for the patch @simongottschlag!

dlorenc (approved):

cpanato:
i did some tests and works well thanks for this!

simongottschlag:
@cpanato no problem at all! 😊👍

PR #234, opened 2022-01-18
Drop SHA1, SHA224 for RSA-PSS/PKCS#1, enforce for RSA-PKCS#1

haydentherapper:
This sets the set of supported hashes to SHA256, SHA384,
and SHA512. SHA1 should not be used for RSA-PKCS#1v1.5,
as it is not collision resistant. Not an expert on this,
but technically SHA1 is acceptable for RSA-PSS due to how
the hash function is used, as part of its
"mask generation function". However, since SHA1 is not
supported by cloud KMS services, we should just remove it entirely.
SHA224 is also not supported anywhere.

RSA-PKCS#1v1.5 was not enforcing the set of hashes on load.

#### Ticket Link

Fixes

#### Release Note

```release-note
Dropped support for SHA1 and SHA224 when signing and verifying with RSA-PKCS#1-v1.5 and RSA-PSS
```

### DIFF
--- a/pkg/signature/rsapkcs1v15.go
+++ b/pkg/signature/rsapkcs1v15.go
@@ -33,13 +33,13 @@ type RSAPKCS1v15Signer struct {
 
 // LoadRSAPKCS1v15Signer calculates signatures using the specified private key and hash algorithm.
 //
-// hf must not be crypto.Hash(0).
+// hf must be either SHA256, SHA388, or SHA512.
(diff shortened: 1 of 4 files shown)


haydentherapper:
cc @lukehinds - Can you kick off the workflows?

haydentherapper:
cc @bobcallaway too, are you able to kick off the workflows?

dlorenc (approved):

PR #246, opened 2022-01-25
Skip strict check on PKCE discovery claim on Azure

bobcallaway:
This skips the check for the `code_challenge_methods_supported` claim
within the OIDC discovery document when configured to connect to an
Azure-based OIDC provider. AFAIK, Azure OIDC does in fact support PKCE
but does not advertise this in its discovery document so this should
enable additional use cases with `cosign` and Azure.

Partially-Fixes: https://github.com/sigstore/cosign/issues/1354

Signed-off-by: Bob Callaway <bob.callaway@gmail.com>

### DIFF
--- a/pkg/oauthflow/pkce.go
+++ b/pkg/oauthflow/pkce.go
@@ -19,6 +19,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
+	"regexp"
 
 	"github.com/coreos/go-oidc/v3/oidc"
 	"golang.org/x/oauth2"
@@ -57,7 +58,7 @@ func NewPKCE(provider *oidc.Provider) (*PKCE, error) {
 			fmt.Printf("Unsupported code challenge method in list: '%v'", method)
 		}
 	}
-	if chosenMethod == "" {
+	if chosenMethod == "" && !providerIsAzureBacked(provider) {
 		return nil, fmt.Errorf("PKCE is not supported by OIDC provider '%v'", provider.Endpoint().AuthURL)
 	}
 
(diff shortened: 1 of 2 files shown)


jw-s:
@bobcallaway I'll build this branch tomorrow and test it against Azure if you need this verification?

bobcallaway:
> @bobcallaway I'll build this branch tomorrow and test it against Azure if you need this verification?

please, that would be awesome

dlorenc (dismissed):

pkg/oauthflow/pkce.go
```diff
@@ -57,7 +58,7 @@ func NewPKCE(provider *oidc.Provider) (*PKCE, error) {
 			fmt.Printf("Unsupported code challenge method in list: '%v'", method)
 		}
 	}
-	if chosenMethod == "" {
+	if chosenMethod == "" && !providerIsAzureBacked(provider) {
```
jw-s, line 61: This means we pass an empty string as the `code_challenge_method` query parameter value which is wrong. This needs to be set to `S256` as per RFC7636 it should support it.

bobcallaway:
@jw-s could you take another look at the updated PR when you have a chance? Thanks!

jw-s:
@bobcallaway  Sorry for the delay - It works as expected now, tested via the interactive flow with confidential client (client secret)

bobcallaway:
> @bobcallaway Sorry for the delay - It works as expected now, tested via the interactive flow with confidential client (client secret)

Thanks for verifying this!

dlorenc (approved):

PR #276, opened 2022-02-15
pkg/signature/kms doesn't depend on kms impls

imjasonh:
#### Summary

Clients who want to enable specific kms implementations can import (or underscore-import) specific KMS impls they want, and otherwise don't have to depend on them.

#### Release Note

```release-note
Specific kms implementations that are needed must be explicitly imported for init-time setup.
```

### DIFF
--- a/pkg/signature/kms/aws/client.go
+++ b/pkg/signature/kms/aws/client.go
@@ -35,7 +35,14 @@ import (
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/pkg/errors"
 	"github.com/sigstore/sigstore/pkg/signature"
+	sigkms "github.com/sigstore/sigstore/pkg/signature/kms"
 )
+
+func init() {
+	sigkms.AddProvider(ReferenceScheme, func(_ context.Context, keyResourceID string, _ crypto.Hash, _ ...signature.RPCOption) (sigkms.SignerVerifier, error) {
+		return LoadSignerVerifier(keyResourceID)
+	})
+}
 
 const (
 	cacheKey = "signer"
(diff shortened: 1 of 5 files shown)


imjasonh:
NB: This constitutes a breaking change, since consumers of `pkg/signature/kms` now must explicitly import the KMS impls they want.

Based on my testing, a client that previously depended on all the KMSes and after this change depends on none of them, drops from 72 MB -> 63 MB (-12.5%)

dlorenc:
> Based on my testing, a client that previously depended on all the KMSes and after this change depends on none of them, drops from 72 MB -> 63 MB (-12.5%)

Nice!

dlorenc (approved):

lukehinds:
@cpanato , note this is a breaking change and will need a major release and some release notes explaining this requires an explicit import. We are long due a release as there are quite few bug fixes and some feature channges since v1.1.0. Do you think we should round up those in 1.2.0 and then plan a 2.0?

cpanato:
@lukehinds @imjasonh @dlorenc i would say we plan for a 2 and check if there are any other changes we need to do.

also Jason, can you add a changelog entry for this? then we don't miss when doing the release

dlorenc:
Note that doing a version 2 bump is a bit tricky with go modules...

I'd argue (mostly out of laziness) that this one isn't actually breaking because the interfaces don't really change.

imjasonh:
> Note that doing a version 2 bump is a bit tricky with go modules...
> 
> I'd argue (mostly out of laziness) that this one isn't actually breaking because the interfaces don't really change.

I agree. Though this represents a behavior change that consumers should be aware of and careful of, I don't think it rises to the level that warrants the toil of cutting and maintaining a v2 of the module.

I'll make sure to call this out in the changelog.

dlorenc:
> Maybe I misunderstood, but the change requires users to update their code or it will break?

This falls into the "it depends" category that makes semver so hard. There are ways you could be calling this today that would break, but it's probably rare.

PR #277, opened 2022-02-17
feat(kms): add supported providers func

Dentrax:
Signed-off-by: Furkan <furkan.turkal@trendyol.com>

Cross ref: https://github.com/sigstore/cosign/pull/1475/files

To provide better UX for the consumer clients to check whether given provider exist or not.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


imjasonh:
FYI: This might have some overlap with https://github.com/sigstore/sigstore/pull/276 -- I'm fine if this gets merged first, in either case one of us will have to rebase.

pkg/signature/kms/kms_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_SupportedProviders(t *testing.T) {
+	supportedProviders := SupportedProviders()
+	require.Contains(t, supportedProviders, aws.ReferenceScheme)
+	require.Contains(t, supportedProviders, azure.ReferenceScheme)
+	require.Contains(t, supportedProviders, gcp.ReferenceScheme)
+	require.Contains(t, supportedProviders, hashivault.ReferenceScheme)
```
imjasonh, line 33: Actually, after https://github.com/sigstore/sigstore/pull/276 this could be an interesting test that KMS impls _aren't_ provided unless imported.

We could remove one of these imports and check that `!require.Contains(..., "gcpkms")`

Dentrax: Oh, good point! I think it would be better to wait your PR to get merged first, and then I can rewrite this test.

dlorenc: This should be ready to rewrite!

imjasonh: I'd try something like this:

```
import (
	"github.com/sigstore/sigstore/pkg/signature/kms/aws"
	"github.com/sigstore/sigstore/pkg/signature/kms/azure"
	// NOT import gcp
	"github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
)

...

require.Contains(t, supportedProviders, aws.ReferenceScheme)
require.Contains(t, supportedProviders, azure.ReferenceScheme)
require.Contains(t, supportedProviders, "gcpkms://") // avoids gcp import, which would register it
require.Contains(t, supportedProviders, hashivault.ReferenceScheme)
```

It's possible the test somehow already depends on `gcp` so this doesn't work, but I think it's worth a shot at least. If it's too difficult to figure out I think just testing the happy path would be fine.

Dentrax: Ah, get it! But I think we should cover all the scheme references by importing here. Otherwise it could be false-positive _pass_ for this unit test if we change the scheme references i.e., hashivault to vault, or if we move the _package_ path.

imjasonh: If you want to be exhaustive you could probably come up with some solution using build tags or requiring manual registration (not in `init()`), I'm just not sure it's worth the maintenance burden.

These url schemes shouldn't change, ever, and if we move a package the test will fail anyway until someone fixes it. I think it's a good compromise to cover the general idea of selective providers. (If it works at all, TBD)

imjasonh:
Awesome it worked! 🎉 

(just kidding, I definitely knew it would work 😅 )

PR #320, opened 2022-03-09
Initial introduction and implementation of `oidc.IDTokenSource`

dekkagaijin:
This PR introduces the `oidc.IDTokenSource` concept and refactors the existing `oauthflow.InteractiveIDTokenGetter` as part of the overall SDK work: https://github.com/sigstore/cosign/issues/844

We want to conceptually isolate the task of retrieving Open ID Tokens, I've introduced the concept of an `IDTokenSource`. This is a prerequisite for implementing a sane Fulcio client for `cosign`, and unblocking higher level work.

```release-note
Introduced `oidc.IDTokenSource` to abstract the generation of Open ID tokens
```


pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,180 @@
...
+	cfg.RedirectURL = oobRedirectURI
+	authURL := cfg.AuthCodeURL(stateToken, opts...)
+	fmt.Fprintln(os.Stderr, "Go to the following link in a browser:\n\n\t", authURL)
+	fmt.Fprintf(os.Stderr, "Enter verification code: ")
+	var code string
+	fmt.Scanln(&code)
+	return code
+}
+
+func startRedirectListener(state, htmlPage string, doneCh chan string, errCh chan error) (*http.Server, *url.URL, error) {
```
mattmoor, line 48: Generally when I see `doneCh` I think `chan struct{}{}`, maybe `codeCh` is a better name?

pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,180 @@
...
+	listener, err := net.Listen("tcp", "localhost:0")
```
mattmoor, line 49: `:0` is probably worth a comment.  I'm guessing it means "pick a port"?

mattmoor, line 49: Hmm, this is never closed 🤔

pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,180 @@
...
+		if r.FormValue("state") != state {
+			errCh <- errors.New("invalid state token")
+			return
+		}
+		doneCh <- r.FormValue("code")
+		fmt.Fprint(w, htmlPage)
+	})
+
+	go func() {
+		if err := s.Serve(listener); err != nil && err != http.ErrServerClosed {
```
mattmoor, line 80: errors.Is?

pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func getCode(doneCh chan string, errCh chan error) (string, error) {
+	timeoutCh := time.NewTimer(120 * time.Second)
+	select {
+	case code := <-doneCh:
+		return code, nil
+	case err := <-errCh:
+		return "", err
+	case <-timeoutCh.C:
```
mattmoor, line 95:
```suggestion
	case <-time.After(120 * time.Second):
```

pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,180 @@
...
+	return ksuid.New().String()
+}
+
+func doInteractiveIDTokenFlow(ctx context.Context, cfg oauth2.Config, p *coreoidc.Provider, extraAuthCodeOpts []oauth2.AuthCodeOption, openBrowser browserOpener) (*IDToken, error) {
+	// generate random fields and save them for comparison after OAuth2 dance
+	stateToken := newKSUID()
+	nonce := newKSUID()
+
+	doneCh := make(chan string)
+	errCh := make(chan error)
```
mattmoor, line 111: Are these ever closed?

mattmoor (commented):
Generally LGTM, but noticed a bunch of random stuff.

We should probably get golangci-lint running here with all the checks we have in cosign

pkg/oauth/oidc/interactive.go
dekkagaijin: It's closed in this perfectly clear block of `doInteractiveIDTokenFlow` :p

```
defer func() {
  go func() {
    _ = redirectServer.Shutdown(context.Background())
  }()
}()
```

pkg/oauth/oidc/interactive.go
dekkagaijin: done

pkg/oauth/oidc/interactive.go
dekkagaijin: done

pkg/oauth/oidc/interactive.go
dekkagaijin: they are now!

dekkagaijin:
> We should probably get golangci-lint running here with all the checks we have in cosign

I'll take a look and see how they differ

dekkagaijin:
PR to update golangci-lint: https://github.com/sigstore/sigstore/pull/322

mattmoor (approved):

pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+type interactiveIDTokenSource struct {
+	cfg               oauth2.Config
+	oidp              *coreoidc.Provider
+	extraAuthCodeOpts []oauth2.AuthCodeOption
+	browser           browserOpener
+}
+
+var errWontOpenBrowser = errors.New("not opening that browser")
```
bobcallaway, line 168: can we make this message a bit more user-friendly? Is there some way that the user can change this (e.g. verify the default system-wide browser is operational?)

bobcallaway (commented):
i only have one nit on an error message, but the bigger question i have is whether the existing e2e tests need to be updated to see that this new package works correctly?

pkg/oauth/oidc/interactive.go
dekkagaijin: This won't ever be surfaced to a user- it only exists to force this flow to use command line inputs over browser redirects

dekkagaijin:
> but the bigger question i have is whether the existing e2e tests need to be updated to see that this new package works correctly?

The existing e2e tests don't exercise any of the code that I moved around to here. I've written tests for this flow before, but I'll need to do a more intrusive refactor in order to make the current code testable.

dekkagaijin:
never mind, found, copied, and adapted the e2e test

pkg/oauth/oidc/interactive.go
bobcallaway: if someone called `InteractiveIDTokenSource` with `allowBrowser=false`, it seems like line 142 would print this message; I could be misreading it though.

pkg/oauth/oidc/interactive.go
```diff
@@ -0,0 +1,185 @@
...
+		return "", errors.New("timeout")
+	}
+}
+
+// newKSUID returns a globally unique, base62 (URL-safe) encoded, 27 character string.
+func newKSUID() string {
+	return ksuid.New().String()
+}
+
+func doInteractiveIDTokenFlow(ctx context.Context, cfg oauth2.Config, p *coreoidc.Provider, extraAuthCodeOpts []oauth2.AuthCodeOption, openBrowser browserOpener) (*IDToken, error) {
```
bobcallaway, line 104: is there a reason not to make this a method on `interactiveIDTokenSource`? all the args except for `ctx` end up being private member variables of the same struct?

dekkagaijin: done

bobcallaway (approved):

PR #326, opened 2022-03-14
return immediately, without waiting for the operation in progress to complete

cpanato:
#### Summary
we should not stream the logs from the cloudbuild to the GitHub action job to avoid any possible leaks. 
so after the job is submitted it returns and completes the task

related to https://github.com/sigstore/fulcio/runs/5536478096?check_suite_focus=true the cloudbuild ran successfully but the svc account does not have permission to stream the logs, and then it failed.

#### Ticket Link
n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/reusable-release.yml
+++ b/.github/workflows/reusable-release.yml
@@ -62,5 +62,5 @@ jobs:
 
       - name: Start cloudbuild job
         working-directory: ./src/github.com/sigstore/${{ inputs.repo }}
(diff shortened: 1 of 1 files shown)


dlorenc (approved):

lukehinds (commented):
LGTM, and thanks for writing a meaningful commit message

lukehinds:
opps, I missed it.

PR #344, opened 2022-03-23
Add method to validate public key

haydentherapper:
This pulls in the validation logic from Cosign, that's currently used to
validate imported keys. I'd like to also validate keys before being
certified by Fulcio.

I switched out the RSA verification logic with a library from Let's
Encrypt. In addition to key size checks, it also checks for small primes
in the modulus, weak primes from Infineon hardware (roca), and easily
factored keys using Fermat's factorization method (which was recently in
the news, a researcher found weak keys generated by printers).

We can't use the ECDSA verification logic because Let's Encrypt doesn't
support the P-521 curve (because browsers don't currently).

#### Ticket Link

Fixes

#### Release Note

```release-note
Added public key validation for RSA and ECDSA keys
```


haydentherapper:
cc @bobcallaway

lukehinds:
Do you also plan to deprecate the same code in cosign and have cosign call into sigstore/sigstore as a library for PK validation?

haydentherapper:
@lukehinds Yep!

lukehinds (dismissed):

dlorenc (approved):

PR #346, opened 2022-03-24
Add method to check for public key equality

haydentherapper:
This is useful when comparing a provided public key with a key from
another source such as a certificate.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added method to check for public key equality
```

### DIFF
--- a/pkg/cryptoutils/publickey.go
+++ b/pkg/cryptoutils/publickey.go
@@ -17,12 +17,17 @@ package cryptoutils
 
 import (
 	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
 	"crypto/sha1" // nolint:gosec
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/asn1"
+	"encoding/hex"
 	"encoding/pem"
 	"errors"
+	"fmt"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


haydentherapper:
cc @dlorenc @bobcallaway

pkg/cryptoutils/publickey.go
```diff
@@ -78,3 +81,24 @@ func SKID(pub crypto.PublicKey) ([]byte, error) {
 	skid := sha1.Sum(spki.SubjectPublicKey.Bytes) // nolint:gosec
 	return skid[:], nil
 }
+
+// EqualKeys compares two public keys. Supports RSA, ECDSA and ED25519.
+func EqualKeys(first, second crypto.PublicKey) error {
+	switch pub := first.(type) {
+	case *rsa.PublicKey:
```
dlorenc, line 88: Is there any way we could do this with interfaces rather than the switch?

haydentherapper: I don't believe so. `crypto.PublicKey` doesn't define `Equals`, only the key types do.

If you mean using the PublicKey interface defined by Sigstore, I don't think it makes sense to use that since we'd have to initialize a verifier for both public keys to be able to compare them.

pkg/cryptoutils/publickey.go
```diff
@@ -78,3 +81,24 @@ func SKID(pub crypto.PublicKey) ([]byte, error) {
...
+		if !pub.Equal(second) {
+			return errors.New("rsa public keys are not equal")
```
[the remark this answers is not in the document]
haydentherapper: What do you think about the base64 encoding or a fingerprint of the key? Maybe reuse SKID from above and log that?

pkg/cryptoutils/publickey.go
dlorenc: Ooof, yeah you're right. This is something generics might actually help with! I keep trying to come up with use cases :)

This looks right for now, thanks.

pkg/cryptoutils/publickey.go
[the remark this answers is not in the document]
lukehinds: subject key identifier sounds good!

pkg/cryptoutils/publickey.go
haydentherapper: Very much agree! I'm also looking forward to getting to use generics once we upgrade to 1.18.

pkg/cryptoutils/publickey.go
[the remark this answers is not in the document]
haydentherapper: Added this!

PR #347, opened 2022-03-25
add oidc callback port flag to use a specific port instead of random (0)

tuananh:
Signed-off-by: Tuan Anh Tran <anhtt17@techcombank.com.vn>

Require another PR at cosign project. https://github.com/sigstore/cosign/pull/1658

#### Summary

Allow to set callback port to a specific port instead of the random one (0) currently. this is a litmitation of some provider that do callback url whitelisting.

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1105

#### Release Note

```release-note

```

### DIFF
--- a/pkg/oauth/oidc/interactive.go
+++ b/pkg/oauth/oidc/interactive.go
@@ -125,7 +125,7 @@ func (idts *interactiveIDTokenSource) IDToken(ctx context.Context) (*IDToken, er
 	codeCh := make(chan string)
 	errCh := make(chan error)
 	// starts listener on ephemeral port
-	redirectServer, redirectURL, err := startRedirectListener(stateToken, oauth.InteractiveSuccessHTML, codeCh, errCh)
+	redirectServer, redirectURL, err := startRedirectListener(stateToken, oauth.InteractiveSuccessHTML, 0, codeCh, errCh)
 	if err != nil {
 		close(codeCh)
 		close(errCh)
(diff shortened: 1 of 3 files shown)


pkg/oauth/oidc/interactive.go
```diff
@@ -45,8 +45,8 @@ func doOobFlow(cfg *oauth2.Config, stateToken string, opts []oauth2.AuthCodeOpti
 	return code
 }
```
lukehinds, line 47: It's a valid case to have a caller set their own port, but this approach will break existing clients who are not providing a port. For them `oidcCallbackPort` will be `nil` which will cause the listen bind to fail.

I would suggest checking if a port value is passed in and if not, then use a random port. 

I expect would also need unit tests for this and the integration tests would need to be changed.

lukehinds (changes requested):

pkg/oauth/oidc/interactive.go
tuananh: i thought default value for int in go is 0. so without the flag, the behavior is same as before.

tuananh: I tested and it works like before.

hectorj2f: @lukehinds I agree we need to keep the default behaviour as it is. I mentioned here (https://github.com/sigstore/cosign/pull/1658#pullrequestreview-921547531) to add a flag with the redirect URI itself instead of just passing the port. There would be situations where localhost must not be ideal, so just exposing the port isn't ideal in my opinion.

lukehinds: @hectorj2f agree, it makes sense we also allow a IP to be passed in. 

> do you mean this is a lib and function signature should not be changed ?

Yes, its used by cosign (iirc) and other libraries.

PR #353, opened 2022-03-28
oidc: set the redirect url if needed

hectorj2f:
#### Summary

When using certain OIDC providers, you might need to set a specific `redirect_uri`. This implementation defaults to the old mechanism setting a listener on localhost and letting the OS to pick a port if the `redirect_uri` is not set.

#### Ticket Link

Fixes https://github.com/sigstore/cosign/issues/1105 and https://github.com/sigstore/cosign/issues/1311

#### Release Note

```release-note
oauthflow: accept a pre-defined redirect_uri.
```

### DIFF
--- a/pkg/oauth/oidc/interactive.go
+++ b/pkg/oauth/oidc/interactive.go
@@ -45,12 +45,12 @@ func doOobFlow(cfg *oauth2.Config, stateToken string, opts []oauth2.AuthCodeOpti
 	return code
 }
 
-func startRedirectListener(state, htmlPage, redirectURI string, codeCh chan string, errCh chan error) (*http.Server, *url.URL, error) {
+func startRedirectListener(state, htmlPage, redirectURL string, codeCh chan string, errCh chan error) (*http.Server, *url.URL, error) {
 	var listener net.Listener
 	var urlListener *url.URL
 	var err error
 
-	if redirectURI == "" {
+	if redirectURL == "" {
 		listener, err = net.Listen("tcp", "localhost:0") // ":0" == OS picks
 		if err != nil {
 			return nil, nil, err
@@ -63,7 +63,7 @@ func startRedirectListener(state, htmlPage, redirectURI string, codeCh chan stri
 			Path:   "/auth/callback",
 		}
 	} else {
-		urlListener, err = url.Parse(redirectURI)
+		urlListener, err = url.Parse(redirectURL)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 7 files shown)


bobcallaway:
Can you give an example of a provider that would require this? I'll take a closer look at the code but if you could help describe the deployment scenario for this I'd appreciate it.

pkg/oauthflow/flow.go
```diff
@@ -94,7 +94,7 @@ var PublicInstanceMicrosoftIDTokenGetter = &InteractiveIDTokenGetter{
 }
 
 // OIDConnect requests an OIDC Identity Token from the specified issuer using the specified client credentials and TokenGetter
```
bobcallaway, line 96: can you add a comment for this function denoting the use of redirectURI (if blank, you get `localhost:0` but otherwise you will get an HTTP listener at the hostname:port specified)?

pkg/oauth/oidc/interactive.go
```diff
@@ -124,8 +138,8 @@ func (idts *interactiveIDTokenSource) IDToken(ctx context.Context) (*IDToken, er
 
 	codeCh := make(chan string)
 	errCh := make(chan error)
-	// starts listener on ephemeral port
-	redirectServer, redirectURL, err := startRedirectListener(stateToken, oauth.InteractiveSuccessHTML, codeCh, errCh)
+	// starts listener using the redirect_uri, otherwise stars on ephemeral port
```
bobcallaway, line 141:
```suggestion
	// starts listener using the redirect_uri, otherwise starts on ephemeral port
```

bobcallaway (commented):
I'd change any use of `URI` to `URL` for both consistency and correctness sake. Other than the other nits mentioned, LGTM. 

Also it may be obvious but if a user specifies a port < 1024 then the calling program will need to run as root (or be granted sufficient privileges) to successfully listen on that port.

pkg/oauthflow/flow.go
hectorj2f: Done :).

hectorj2f:
@bobcallaway Thanks for the review.

> Also it may be obvious but if a user specifies a port < 1024 then the calling program will need to run as root (or be granted sufficient privileges) to successfully listen on that port.

I agree with you here, but we cannot avoid scenarios where the user sets `--oidc-redirect-url=http://localhost:0/myideal/path/callbakc` if they want. I know it is not a common scenario.

bobcallaway (approved):
thanks!

PR #361, opened 2022-03-31
Add fake signer that implements KMS interface

haydentherapper:
This allows us to test code that uses kms.Get by importing the fake KMS
package.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added fake KMS signer for testing
```

### DIFF
--- a/pkg/signature/kms/fake/signer.go
+++ b/pkg/signature/kms/fake/signer.go
@@ -1,0 +1,134 @@
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


haydentherapper:
cc @bobcallaway

hectorj2f (approved):
Lgtm, and it would be useful when adding some tests in cosigned

dlorenc (approved):

haydentherapper:
Need to make one quick change before this is submitted, I want to provide a way to pass in a private key to the loader. Otherwise if you call kms.Get() repeatedly, you'll get back different key values each time.

haydentherapper:
Added: A bit of a weird way to do it, but I'm passing a marshalled ECDSA private key via the string ref. If the string ref after `fakekms://` is empty, then it'll generate a key for you.

haydentherapper:
You know what, it'd be far easier to just pass the private key in the context. I'll clean this up.

haydentherapper:
Ready for review!

dlorenc (approved):

PR #419, opened 2022-04-30
Call ValidReference in all KMS cases

imjasonh:
These methods were defined in all the KMS impls, but were only actually
called in the GCP case. It's possible they were used outside this
pacekage, since they're exported, but for consistency we should ensure
the reference is valid before proceeding to make API calls.

Signed-off-by: Jason Hall <jason@chainguard.dev>

```release-note
KMS implementations all ensure the reference is valid before proceeding.
```


lukehinds (approved):
Nice catch!

PR #430, opened 2022-05-09
Add method to unmarshal certificates with a limit

haydentherapper:
This removes a DOS vector for services that use this method. Otherwise,
a client can provide a large PEM block to cause the service to do a
significant amount of work.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added method to unmarshal certificates from a PEM block with a limit
```

### DIFF
--- a/pkg/cryptoutils/certificate.go
+++ b/pkg/cryptoutils/certificate.go
@@ -76,6 +76,35 @@ func UnmarshalCertificatesFromPEM(pemBytes []byte) ([]*x509.Certificate, error) 
 	return result, nil
 }
 
+// UnmarshalCertificatesFromPEMLimited extracts one or more X509 certificates from the provided
+// byte slice, which is assumed to be in PEM-encoded format. Fails after a specified
+// number of iterations.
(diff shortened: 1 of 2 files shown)


haydentherapper:
cc @dlorenc @bobcallaway

pkg/cryptoutils/certificate.go
```diff
@@ -76,6 +76,35 @@ func UnmarshalCertificatesFromPEM(pemBytes []byte) ([]*x509.Certificate, error)
 	return result, nil
 }
 
+// UnmarshalCertificatesFromPEMLimited extracts one or more X509 certificates from the provided
+// byte slice, which is assumed to be in PEM-encoded format. Fails after a specified
+// number of iterations.
+func UnmarshalCertificatesFromPEMLimited(pemBytes []byte, iterations int) ([]*x509.Certificate, error) {
```
dlorenc, line 82: Nice! Should we suggest a safe/reasonable limit?

haydentherapper: Added a suggestion of 10! It'd be quite unexpected to see a chain longer than that.

dlorenc (approved):

PR #435, opened 2022-05-11
Move fulcioroots and tuf packages from cosign

imjasonh:
#### Summary

This moves these packages from sigstore/cosign into sigstore/sigstore.

- `pkg/fulcioroots` comes from cosign's [`cmd/cosign/cli/fulcio/fulcioroots@2ccdb3`](https://github.com/sigstore/cosign/tree/2ccdb3e573eaf8839838e88a7dcd09f956809dbd/cmd/cosign/cli/fulcio/fulcioroots), and drops that package's behavior when the `SIGSTORE_ROOT_FILE` env var is set -- this will remain in `sigstore/cosign`.
- `pkg/tuf` comes from cosign's [`pkg/cosign/tuf@2ccdb3`](https://github.com/sigstore/cosign/tree/2ccdb3e573eaf8839838e88a7dcd09f956809dbd/pkg/cosign/tuf) and is otherwise largely unchanged. Some methods were unexported that aren't used outside of this package.

Part of https://github.com/sigstore/cosign/issues/1865

#### Release Note

```release-note
pkg/fulcioroots and pkg/tuf are moved from cosign repo
```


imjasonh:
The lint check is failing due to un-doc-commented exported methods, but I think I'd rather just leave those as-is than balloon the changes in this PR. If folks feel strongly about having these commented I think we should do that in a separate PR.

imjasonh:
@haydentherapper @asraa

pkg/fulcioroots/fulcioroots.go
```diff
@@ -0,0 +1,154 @@
...
+0N/RgBJz/9jWCiXno3sweTAOBgNVHQ8BAf8EBAMCAQYwEwYDVR0lBAwwCgYIKwYB
+BQUHAwMwEgYDVR0TAQH/BAgwBgEB/wIBADAdBgNVHQ4EFgQU39Ppz1YkEZb5qNjp
+KFWixi4YZD8wHwYDVR0jBBgwFoAUWMAeX5FFpWapesyQoZMi0CrFxfowCgYIKoZI
+zj0EAwMDZwAwZAIwPCsQK4DYiZYDPIaDi5HFKnfxXx6ASSVmERfsynYBiX2X6SJR
+nZU84/9DZdnFvvxmAjBOt6QpBlc4J/0DxvkTCqpclvziL6BCCPnjdlIB3Pu3BxsP
+mygUY7Ii2zbdCdliiow=
+-----END CERTIFICATE-----`
+
+const (
+	altRoot = "SIGSTORE_ROOT_FILE"
```
haydentherapper, line 60: We should also make sure that there's documentation in Cosign somewhere for setting this flag for testing.

pkg/fulcioroots/fulcioroots.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+func Get() *x509.CertPool {
+	rootsOnce.Do(func() {
+		var err error
+		roots, intermediates, err = initRoots()
+		if err != nil {
+			panic(err)
```
haydentherapper, line 68: Panic was fine in the CLI. Given this may be used by services, we should remove the panic and return an error instead. What do you think?

pkg/tuf/policy.go
```diff
@@ -0,0 +1,208 @@
...
+	"encoding/json"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+
+	cjson "github.com/secure-systems-lab/go-securesystemslib/cjson"
+)
+
+type Signed struct {
```
haydentherapper, line 34: I'm fine with adding comments later, but if they could be done immediately in another PR, that'd be great so we don't leave tech debt.

pkg/tuf/signer.go
```diff
@@ -0,0 +1,50 @@
...
+
+package tuf
+
+import (
+	"encoding/json"
+)
+
+const (
+	KeyTypeFulcio   = "sigstore-oidc"
+	KeySchemeFulcio = "https://fulcio.sigstore.dev"
```
haydentherapper, line 24: Does anyone recall what the purpose of this is? Is this a problem that this is hardcoded in, such as for staging?

@dlorenc @bobcallaway maybe?

pkg/fulcioroots/fulcioroots.go
imjasonh: This is definitely already used by services, including the webhook, AFAIK.

Would you accept a TODO for now, and I'll get it in the next PR? I'm trying to keep this as functionally equivalent to previous code.

pkg/tuf/policy.go
imjasonh: I'm not sure I'm the person you want writing doc comments for these, is the only problem 😆 

Could I file an issue (on whom I don't know) to followup and add these comments?

pkg/tuf/signer.go
imjasonh: It doesn't appear to be referenced anywhere within cosign at least, I'm not sure if it was just exported on accident.

pkg/fulcioroots/fulcioroots.go
haydentherapper: That's fine, if you want to assign me to both this and the comments issue, I can handle both right after this is in.

pkg/tuf/policy.go
haydentherapper: I can take care of it!

pkg/tuf/signer.go
haydentherapper: Seems to be used with https://github.com/sigstore/cosign/blob/d2d74640720b40caa335346566fcdbc978f57bba/cmd/cosign/cli/policy_init.go#L77

Scheme isn't referenced though. Probably fine to leave.

haydentherapper (approved):

go.mod
```diff
@@ -33,6 +33,11 @@ require (
 	gopkg.in/square/go-jose.v2 v2.6.0
 )
 
+require (
+	cloud.google.com/go/storage v1.10.0
```
haydentherapper, line 37: Were there concerns about pulling in this library? Was it just due to size?

imjasonh: Yeah I'd like to replace it with regular HTTP calls instead of taking a dependency on this GCS package. It's not _huge_, it's just mostly unnecessary.

pkg/tuf/signer.go
imjasonh: Moved these to unexported in sigstore/sigstore. 👍

imjasonh:
Any more concerns here? If not I can merge and un-WIP https://github.com/sigstore/cosign/pull/1866 to start using this.

haydentherapper:
This is a large pr, I want to double check nothing has changed. I don’t have a lot of confidence that a break would be detected once this is used in cosign due to the lack of tests. 
I’ll take a look later today or early Monday.

haydentherapper:
This LGTM - If you're able to wait a few days, it'd be nice to also get @asraa's LGTM since she's the author of the TUF changes.

pkg/fulcioroots/fulcioroots.go
```diff
@@ -0,0 +1,113 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
asraa, line 2: nit: 2022

pkg/tuf/repository/root.json
```diff
@@ -0,0 +1,144 @@
...
+			},
+			"timestamp": {
+				"keyids": [
+					"b6710623a30c010738e64c5209d367df1c0a18cf90e6ab5292fb01680f83453d"
+				],
+				"threshold": 1
+			}
+		},
+		"spec_version": "1.0",
+		"version": 2
```
asraa, line 142: should we update the embedded repo to root 3?
https://github.com/sigstore/root-signing/tree/main/repository/repository

pkg/tuf/client.go
```diff
@@ -0,0 +1,618 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/theupdateframework/go-tuf/client"
+	tuf_leveldbstore "github.com/theupdateframework/go-tuf/client/leveldbstore"
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/util"
+)
+
+const (
```
asraa, line 42: I think these defaults need to be set in cosign, and stuff in sigstore pkg/ should be agnostic of TUF locations

I think that means that `Initialize` and `NewFromEnv` should be in cosign, because their internal callers resolve the defaults.

imjasonh: How should a non-cosign client get the fulcio roots in that case? Is there some other better way to initialize a TUF client that can get passed to sigstore's `fulcioroots.Get` and `fulcioroots.GetIntermediates`, that doesn't rely on cosign-specific defaults?

(I had assumed that cosign's defaults should become sigstore's defaults, and trickle down to anything interested in getting Fulcio's roots. Cosign had extra behavior based on the presence of the `COSIGN_ROOT_FILE` env var, which was moved back into cosign for now)

pkg/tuf/repository/root.json
haydentherapper: Let's add that task to https://github.com/sigstore/sigstore/issues/436 - I think we should aim to have this PR just be a move

pkg/tuf/client.go
```diff
@@ -0,0 +1,618 @@
...
+	DefaultRemoteRoot = "sigstore-tuf-root"
```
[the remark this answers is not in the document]
haydentherapper: The GCS dependency wasn't dropped yet, so can we punt this task to later?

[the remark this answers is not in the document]
imjasonh: That sounds great, what's the correct HTTP URL to set here, instead of this? Is it just `https://sigstore-tuf-root.storage.googleapis.com`?

And if we do that, should I remove the GCS path in `remoteFromMirror`, and make callers who want to use GCS provide their own TUF client somehow? (Also fine making this work for a future PR, if we want this PR to mainly be code movement with not much functional change)

pkg/tuf/repository/root.json
imjasonh: Added to #436

haydentherapper:
There's active debugging on the TUF client from a recent postmortem, so we may want to hold off on submitting this until that's resolved, as we'd like to get the bug fixed before 1.0

pkg/tuf/client.go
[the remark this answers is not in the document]
asraa: > GCS path in remoteFromMirror, and make callers who want to use GCS provide their own TUF client somehow?

Sure, maybe in the follow-up we should expose the remote as a tuf client.RemoteStore in the API for `Initialize` -- that way library callers can provide their open GCS/other remote store.

imjasonh:
Sounds good, I've converted to a draft for now. After the issue is resolved I can pick it back up.

znewman01:
After we do this: should probably consider https://github.com/sigstore/cosign/issues/1935

imjasonh:
> After we do this: should probably consider [sigstore/cosign#1935](https://github.com/sigstore/cosign/issues/1935)

I'd be fine cleaning it up first then moving it, too. There's no real rush to move it, especially if it causes confusion during cleanups or bug fixes.

imjasonh:
I'd like to pick this up again after the TUF changes that blocked us last time, and after https://github.com/sigstore/cosign/pull/1967 -- let me know if there's anything else I should wait for, otherwise I'll update this PR.

asraa:
> I'd like to pick this up again after the TUF changes that blocked us last time, and after https://github.com/sigstore/cosign/pull/1967 -- let me know if there's anything else I should wait for, otherwise I'll update this PR.

Thanks Jason! I think we're ready for updating this. I'll take a close review for this!

haydentherapper:
Can you move this in a way that preserves commit history? I’ve seen this done for other large moves like this.

imjasonh:
> Can you move this in a way that preserves commit history? I’ve seen this done for other large moves like this.

I probably _can_, but it'll be gross, and I don't really trust myself to do it cleanly. I know it's painful to track history through a move like this, I can add links in the PR description to the exact commit the directories were copied from.

If the concern is edits made during the move that don't show up in a nice diff, that's a reasonable concern too. GitHub won't let me attach diffs (weirdly) but I've put them in a Gist here: https://gist.github.com/imjasonh/3befe1aceca12c0d252623d731073658

haydentherapper:
It’s about tracking how files have changed over time. I think @vaikas is doing this now for the policy controller, do you have suggestions on how to do this?

vaikas:
I'm far from git expert, but i think what i was trying to do was a butt simpler because I was trying to preserve the entire history for the whole repo and not just a subset of files. I have no idea how to do this 😰

znewman01 (dismissed):

wlynch:
@imjasonh I followed a modified version of https://johno.com/move-directory-between-repos-with-git-history for #486 -

```sh
$ cd /tmp
$ git clone https://github.com/sigstore/cosign
$ git filter-branch --subdirectory-filter pkg/providers -- --all
$ cd $GOPATH/src/github.com/sigstore/sigstore
$ git remote add cosign-local /tmp/cosign
$ git checkout cosign-local/main
$ git rebase --signoff main
```

From here move the files to the right files, fix any build / lint issues, and push it up!

imjasonh:
Thanks, that's helpful. Between `gitsign` popping up a browser for every commit and habitually forgetting `--signoff` (both my own fault) this is going slowly, but I think it'll work out. Stay tuned for this PR to become a 50-commit frankenstein! 👹

imjasonh:
> Works for me, but is there any way to disable the linter on this code for now so we don't break HEAD?

I've updated the lint action to only complain about findings in changed lines, that will at least let other PRs get merged while these are un-godoc'd.

We'll still need someone to force-merge this over the complaints of the linter in this change though. Or make the check not-required.

znewman01:
Can you exclude whole directories/specific files instead? IMO that makes more sense,  then we can remove that restriction in a follow-up PR and doesn't require any kind of override.

znewman01 (approved):

haydentherapper:
Ooo today, will take a look tomorrow!

asraa (approved):

dlorenc:
Here we goooo!

PR #441, opened 2022-05-12
Add unsafe verifier to verify signatures with SHA1 digests

haydentherapper:
I relaxed the hash function constraints on VerifyMessage, including the
SHA1 digest as a supported function. The expectation is that
LoadVerifier will still be the primary way to set up a verifier, which
will enforce the hash function. Otherwise, LoadUnsafeVerifier will be
used to load a verifier that only supports SHA1.

Note that SignMessage will not support SHA1 still.

I also dropped SHA1 from ECDSA's supported hash functions.

#### Ticket Link

Fixes

#### Release Note

```release-note
Added support for an unsafe verifier to verify signatures that use SHA1. Only use if SHA1 is required.
```

### DIFF
--- a/pkg/signature/ecdsa.go
+++ b/pkg/signature/ecdsa.go
@@ -26,7 +26,16 @@ import (
 	"github.com/sigstore/sigstore/pkg/signature/options"
 )
 
+// checked on LoadSigner, LoadVerifier and SignMessage
 var ecdsaSupportedHashFuncs = []crypto.Hash{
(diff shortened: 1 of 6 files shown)


haydentherapper:
Unblocks https://github.com/sigstore/rekor/pull/823

haydentherapper:
cc @dlorenc

dlorenc (approved):
Nice!

PR #466, opened 2022-05-25
Export providerInit type

imjasonh:
Also remove unnecessary providerMux indirection, and just use a
package-level var directly.

Signed-off-by: Jason Hall <jason@chainguard.dev>

cc @mattmoor 

---

It's generally not a best practice to have an exported method accept or return unexported types, and exporting this type alias can make code using it cleaner.

```release-note
NONE
```


mattmoor (approved):
Awesome!

bobcallaway (approved):

dlorenc (approved):

PR #468, opened 2022-05-25
Allow passing options to GCP's `LoadSignVerifier`.

mattmoor:
This lets the caller control authentication, in particular by providing an `option.TokenSource`.

#### Release Note
```release-note
gcp.LoadSignerVerifier can now be passed GRPC ClientOptions
```

cc @imjasonh @dlorenc

### DIFF
--- a/pkg/signature/kms/gcp/client.go
+++ b/pkg/signature/kms/gcp/client.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	gcpkms "cloud.google.com/go/kms/apiv1"
+	"google.golang.org/api/option"
 	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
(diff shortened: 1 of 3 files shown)


imjasonh (approved):
Nice.

mattmoor:
cc @bobcallaway @lukehinds 

Not sure who has approvers here, but surely you do 😂

dlorenc (approved):

mattmoor:
@bobcallaway perhaps, but I'm not optimistic there's a canonical option type for all providers :(

AWS is next on my list to play with, and based on my experience with their SDKs there is unlikely to be a compatible option type that both would take.  One pattern I've seen in the past (and don't love) is to pass `...interface{}` and then type switch on the options when you reach a typed interface.  This is most common when something takes a list of two types of options.

Basically what I was playing with (@imjasonh idea) imports the existing stuff, and re-registers the GCP auth provider with additional options directly (with current logic, last registration for a scheme wins).

PR #475, opened 2022-05-28
Migrate AWK KMS to use the v2 SDK.

mattmoor:
Looking at doing something similar to https://github.com/sigstore/sigstore/pull/468 for AWS, I noticed that our KMS stuff was using the old SDK.  The bulk of this change is migrating things to the v2 SDK, but this also exposes a way
to plumb through options to `LoadSignerVerified` similar to #468 for GCP.

#### Release Note

```release-note
The AWS KMS now uses the v2 SDK, and exposes a way to pass AWS SDK options.
```

cc @bobcallaway @imjasonh This sort of highlights the unfortunately different option type exposed by the AWS SDK (vs. something more portable like `oauth2` with GCP 😞 )


dlorenc (approved):
I'm not an AWS expert but this looks smaller than I would have guessed.

bobcallaway (approved):

PR #487, opened 2022-06-03
Add a warning when using WithDigest with ECDSA

### LINKED ISSUE #479: Enforcement of a digest looking like a digest
**Description**

There's an interesting forgery for ECDSA where it's possible to forge a valid signature over a random value for a fixed public key. To defend against this, it's necessary to first hash a message before signing or verifying it. For ED25519, this is handled via a pre-hash, while for ECDSA, it's standard practice to first hash the message.

However, for the hashedrekord type, it's a requirement that we verify against a digest without hashing again. This is why we support `WithDigest`. This makes `WithDigest` unsafe. We can help defend against this if there is a check beforehand that the digest look like a digest, so a random value isn't accepted. For example, Rekor [checks](https://github.com/sigstore/rekor/blob/main/pkg/types/hashedrekord/v0.0.1/entry.go#L171-L173) that the digest matches a SHA256 regex.

This is not a full proof approach, as it is still possible that a random value look like a digest, it's just hard to find such a value for this forgery (I think this is true, but I'd need someone more well-versed in elliptic curve crypto to verify this).

We should add a) warnings in comments about the dangers of verifying with a digest, and b) move the rekor checks into here to enforce that a digest looks like a digest.

haydentherapper:

#### Ticket Link

Fixes #479

#### Release Note

```release-note

```

### DIFF
--- a/pkg/signature/options/digest.go
+++ b/pkg/signature/options/digest.go
@@ -27,6 +27,9 @@ func (r RequestDigest) ApplyDigest(digest *[]byte) {
 }
 
 // WithDigest specifies that the given digest can be used by underlying signature implementations
+// WARNING: When verifying a digest with ECDSA, it is trivial to craft a valid signature
+// over a random message given a public key. Do not use this unles you understand the
+// implications and do not need to protect against malleability.
 func WithDigest(digest []byte) RequestDigest {
 	return RequestDigest{digest: digest}
 }


haydentherapper:
cc @znewman01

cpanato (approved):

hectorj2f (approved):

PR #500, opened 2022-06-13
tuf: add a method to retrieve rekor public keys

asraa:

#### Summary

* Adds a wrapper method to get rekor public keys, to use in cosign

I expect that with this method, we can now set sign/verify opts to include RekorPubKeys, so that people can define this themselves when they cosign as a library. This is how the fulcio roots works as well.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/pkg/cryptoutils/publickey.go
+++ b/pkg/cryptoutils/publickey.go
@@ -55,6 +55,19 @@ func UnmarshalPEMToPublicKey(pemBytes []byte) (crypto.PublicKey, error) {
 	return x509.ParsePKIXPublicKey(derBytes.Bytes)
 }
 
+// UnmarshalPEMToECDSAKey converts a PEM-encoded byte slice into an *ecdsa.PublicKey.
+func UnmarshalPEMToECDSAKey(pemBytes []byte) (*ecdsa.PublicKey, error) {
+	pub, err := UnmarshalPEMToPublicKey(pemBytes)
+	if err != nil {
+		return nil, err
+	}
+	ecdsaPub, ok := pub.(*ecdsa.PublicKey)
+	if !ok {
+		return nil, fmt.Errorf("invalid public key: was %T, require *ecdsa.PublicKey", pub)
+	}
+	return ecdsaPub, nil
+}
+
 // MarshalPublicKeyToDER converts a crypto.PublicKey into a PKIX, ASN.1 DER byte slice
 func MarshalPublicKeyToDER(pub crypto.PublicKey) ([]byte, error) {
 	if pub == nil {
--- a/pkg/rekorpubs/rekorpubs.go
+++ b/pkg/rekorpubs/rekorpubs.go
@@ -1,0 +1,89 @@
+//
+// Copyright 2022 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


pkg/rekorpubs/rekorpubs.go
```diff
@@ -0,0 +1,89 @@
...
+			singletonRekorErr = fmt.Errorf("initializing tuf: %w", err)
+			return
+		}
+		targets, err := tufClient.GetTargetsByMeta(tuf.Rekor, []string{rekorTargetStr})
+		if err != nil {
+			singletonRekorErr = fmt.Errorf("error getting targets: %w", err)
+			return
+		}
+		for _, t := range targets {
+			rekorPubKey, err := cryptoutils.UnmarshalPEMToECDSAKey(t.Target)
```
haydentherapper, line 74: Would `UnmarshalPEMToPublicKey` work, or do we make more assumptions that the Rekor key is an ECDSA key?

asraa: Wow, I lost track of this comment -- yes! The methods that use Rekor public keys expect ecdsa keys: https://github.com/sigstore/cosign/blob/890cec1f43a8ec0754e0dd5a5d120847b63b6c4e/pkg/cosign/verify.go#L818

asraa:
ping @haydentherapper or maybe this should just go in an internal cosign package?

haydentherapper:
I'd prefer this either live in Rekor or sigstore/sigstore. Given it's intertwined with TUF, I think sigstore/sigstore in the best option for now. I think tlog verification should live in Rekor, certificate verification in Fulcio, TUF, crypto/signature/PEM/etc functions in sigstore/sigstore, OCI in Cosign.

pkg/rekorpubs/rekorpubs.go
haydentherapper: Just curious, long-term, how hard would it be to remove the ECDSA requirement?

pkg/cryptoutils/publickey.go
```diff
@@ -55,6 +55,19 @@ func UnmarshalPEMToPublicKey(pemBytes []byte) (crypto.PublicKey, error) {
 	return x509.ParsePKIXPublicKey(derBytes.Bytes)
 }
 
+// UnmarshalPEMToECDSAKey converts a PEM-encoded byte slice into an *ecdsa.PublicKey.
+func UnmarshalPEMToECDSAKey(pemBytes []byte) (*ecdsa.PublicKey, error) {
```
haydentherapper, line 59: Do you think we should make this a private function of rekorpubs? I'd prefer to encourage package users to use the more generic PEM to crypto.PublicKey method.

pkg/rekorpubs/rekorpubs.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return "", err
+	}
+	digest := sha256.Sum256(pubBytes)
+	return hex.EncodeToString(digest[:]), nil
+}
+
+// GetRekorPubs returns a map of rekor public keys keyed by rekor log ID. Each key contains
+// the ecdsa public key of the log and the status of the log (e.g. Active, Inactive).
+func GetRekorPubs() (map[string]RekorPubKey, error) {
```
haydentherapper, line 60: We should also document that this caches the public keys. This could be an issue if used by a server that never restarts and never would fetch the latest keys for example.

Do you think we should have two functions, one for caching and one that doesn't cache?

haydentherapper (changes requested):

haydentherapper:
@asraa Good to close in favor of sigstore-go?

asraa (commented):
Yeah, I guess it's not really quite needed yet anyway. sigstore-go will use TUF to populate the trust root proto

PR #514, opened 2022-06-20
oauthflow/interacive: Make input/output configurable.

wlynch:
#### Summary

This replaces fmt calls outputting to stderr and reading from stdin to
instead act on configurable io.Writer/Reader that default to the same
locations.

This allows other libraries that can't interact with stderr/stdin
directly (like gitsign) to inject its own I/O to the interactive flow.

Also removes unused MessageWriter (which serves a similar purpose as
Output)

#### Ticket Link

Part of https://github.com/sigstore/gitsign/issues/20

#### Release Note

```release-note
Interactive flow library can use configurable I/O sources.
```


dlorenc (approved):

pkg/oauthflow/flow.go
```diff
@@ -57,32 +56,36 @@ func ConnectorIDOpt(prov string) oauth2.AuthCodeOption {
 // DefaultIDTokenGetter is the default implementation.
 // The HTML page and message printed to the terminal can be customized.
 var DefaultIDTokenGetter = &InteractiveIDTokenGetter{
-	MessagePrinter: func(url string) { fmt.Fprintf(os.Stderr, "Your browser will now be opened to:\n%s\n", url) },
-	HTMLPage:       soauth.InteractiveSuccessHTML,
+	HTMLPage: soauth.InteractiveSuccessHTML,
+	Input:    os.Stdin,
```
imjasonh, line 60: Can we make it so a `nil` input falls back to `os.Stdin`? Then each of the token getter vars below doesn't have to set it, and any new one doesn't have to remember to set it.

imjasonh (dismissed):
Nice.

pkg/oauthflow/flow.go
wlynch: Done!

dlorenc (approved):

bobcallaway (approved):

PR #522, opened 2022-06-22
Keep calm and don't panic: enable and apply forcetypeassert lint rules

Dentrax:
Signed-off-by: Furkan <furkan.turkal@trendyol.com>

#### Summary
I was playing with Vault KMS and hit two different panic cases. And both of are actually casting error:
1. [keysData.(map[string]interface{})](https://github.com/sigstore/sigstore/blob/9dde75d80cf82ad42401b08165151d947cdcd538/pkg/signature/kms/hashivault/client.go#L208)
2. [latestVersion.(json.Number)](https://github.com/sigstore/sigstore/blob/9dde75d80cf82ad42401b08165151d947cdcd538/pkg/signature/kms/hashivault/client.go#L213)

I'm not so sure which cases caused this problem. So I thought It would be nice to have a [forcetypeassert](https://github.com/gostaticanalysis/forcetypeassert) analyzer in the golangci-lint.

I also had to remove the following line because forcetypeassert still complains about: `type assertion must be checked`

```
return fmt.Errorf("data type assertion for field `valid` failed: %T %#v", valid.(bool), valid.(bool))
```

I could have passed a `//nolinter:` but I thought in both cases the following above throws panic, right? @blz-ea 

#### Ticket Link

Fixes

#### Release Note

```release-note
Enable forcetypeassert linter in golangic-lint
```

### DIFF
--- a/pkg/signature/kms/hashivault/client.go
+++ b/pkg/signature/kms/hashivault/client.go
@@ -210,18 +210,32 @@ func (h *hashivaultClient) fetchPublicKey(_ context.Context) (crypto.PublicKey, 
 		return nil, errors.New("failed to read transit key keys: Invalid keys map")
 	}
 
-	keyVersion := latestVersion.(json.Number)
+	keyVersion, ok := latestVersion.(json.Number)
+	if !ok {
+		return nil, fmt.Errorf("format of 'latest_version' is not json.Number")
+	}
+
 	keyData, ok := keys[string(keyVersion)]
 	if !ok {
 		return nil, errors.New("failed to read transit key keys: corrupted response")
 	}
 
-	publicKeyPem, ok := keyData.(map[string]interface{})["public_key"]
+	keyMap, ok := keyData.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("could not parse transit key keys data as map[string]interface{}")
+	}
+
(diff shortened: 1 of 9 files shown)


pkg/signature/kms/hashivault/client.go
```diff
@@ -309,7 +323,7 @@ func (h hashivaultClient) verify(sig, digest []byte, alg crypto.Hash, opts ...si
 
 	isValid, ok := valid.(bool)
 	if !ok {
-		return fmt.Errorf("data type assertion for field `valid` failed: %T %#v", valid.(bool), valid.(bool))
+		return fmt.Errorf("received non-bool value from 'valid' key")
```
dlorenc, line 326: should we use something like %T to still get the type or %s to attempt to stringify it?

Dentrax: Reproduced the same case on playground, and it threw panic for the `valid.(bool)` casting: https://go.dev/play/p/-rCmjQsK0e8

But works well if we just use

```go
fmt.Printf("%T %#v", valid, valid)
```

I can change all type assert errors to get their types `%T` and stringify `%#v` but I'm not sure if they are safe to print the entire content to stdout in vault client. 🤔

dlorenc (approved):

PR #536, opened 2022-06-29
Remove err from type cast failures

mtrmac:
#### Summary
A cleanup for #522: Remove err from type cast failures

In both cases the error value is unrelated to the cast failure, and in fact guaranteed to be `nil` on that code path.

#### Release Note
```release-note
NONE
```


cpanato (approved):
Thanks!

bobcallaway (approved):

PR #554, opened 2022-07-11
tuf: remove test that targets list is complete

asraa:

#### Summary

* Removes requiring the target list is complete in the TUF client test. This list of targets can change during root signing events, and this test shouldn't be so fragile. Only keeps stable targets.


cpanato (approved):

asraa:
looking for one more review? @bobcallaway @haydentherapper

haydentherapper (approved):

bobcallaway (approved):

Issue #211, opened 2021-12-24
Include code coverage for tests


naveensrinivasan:
**Description**
Right now the tests don't have coverage metrics reported. It would help when including a new feature to identify the coverage.

naveensrinivasan:
I am planning to work on this.

naveensrinivasan:
### Coverage tool

- Coverage threshold should be configurable for each repository - for example, `70%` within the repository.
- A setting file should be able to override the coverage threshold for a given package within the repository. `[github.com/foo/bar/xyz] : 61`
- The tool should use native `go` tools and not depend on external vendors.
- The tool should be configurable as part of the PR to fail if the desired threshold is not met.
- Contributors should be able to run it locally if desired before doing a PR.

haydentherapper:
As discussed on issue, codecov now supports this.

PR #587, opened 2022-07-30
New : Ensure code coverage is met

naveensrinivasan:
#### Summary
- This will enable the desired code coverage to be met for the project
- The coverage is set to 70 to start with. This setting is in the `Makefile`. 
- The `coveragethreshold.json` is an override for packages with different coverage needs from the global coverage threshold.
For example `github.com/sigstore/sigstore/pkg/oauthflow` package has only coverage of `36.4` percentage. The tool would ensure for the `github.com/sigstore/sigstore/pkg/oauthflow` the coverage doesn't fall below `36.4`
- The code coverage tool uses the standard go tool.
-  This PR will not break on the existing package coverage. Only new code changes if they fall below the desired coverage requirements will fail.
- https://github.com/sigstore/sigstore/issues/211

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all pkg test test-e2e clean lint fuzz help
+.PHONY: all pkg test test-e2e clean lint fuzz help coverage
 
-all: pkg fuzz
+all: pkg fuzz coverage
 
 TOOLS_DIR := hack/tools
 TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/bin)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -24,6 +24,8 @@ INTEGRATION_TEST_DIR := ./test/e2e
 GO-FUZZ-BUILD := $(TOOLS_BIN_DIR)/go-fuzz-build
 GENSRC = pkg/generated/models/%.go pkg/generated/client/%.go
 SRCS = $(shell find pkg -iname "*.go"|grep -v pkg/generated) $(GENSRC)
+TEST_COVERAGE_PERCENTAGE = 70
```
naveensrinivasan, line 27: I have set the global test coverage to be `70%`. I didn't know what to start with. Open to what the project wants it to be.

dlorenc:
In the past we decided against enforcing arbitrary limits, and instead monitoring it over time.

cc @lukehinds

haydentherapper:
I've also found the coverage calculations to be inaccurate at times.

haydentherapper:
We should be able to just use codecov - https://docs.codecov.com/docs/commit-status

haydentherapper:
Going to close this out, let’s pursue codecov if we want to enforce code coverage

Issue #600, opened 2022-08-07
StaticTokenGetter not working without network connection.


vaikas:
**Description**
While working on a plane without wifi, noticed that the following was unexpectedly failing, since I was running scaffolding on a local cluster and using OIDC tokens off the kind cluster. For example, a sign command was failing because it couldn't talk to oauth2.sigstore.dev:

```
vaikas@villes-mbp scaffolding % COSIGN_EXPERIMENTAL=1 ../cosign/main sign --rekor-url $REKOR_URL --fulcio-url $FULCIO_URL --force --allow-insecure-registry $demoimage --identity-token `curl -s $ISSUER_URL`

Generating ephemeral keys...
Retrieving signed certificate...

        Note that there may be personally identifiable information associated with this signed artifact.
        This may include the email address associated with the account with which you authenticate.
        This information will be used for signing this artifact and will be stored in public transparency logs and cannot be removed later.
Error: signing [registry.local:5000/knative/demo@sha256:b6cfc6e87706304be13f607b238d905db1096619c0217c82f4151117e0112025]: getting signer: getting key from Fulcio: retrieving cert: Get "https://oauth2.sigstore.dev/auth/.well-known/openid-configuration": dial tcp: lookup oauth2.sigstore.dev: no such host
main.go:62: error during command execution: signing [registry.local:5000/knative/demo@sha256:b6cfc6e87706304be13f607b238d905db1096619c0217c82f4151117e0112025]: getting signer: getting key from Fulcio: retrieving cert: Get "https://oauth2.sigstore.dev/auth/.well-known/openid-configuration": dial tcp: lookup oauth2.sigstore.dev: no such host
```

Where my identity-token fetch goes to local kind cluster to fetch an oidc token.

So, we unnecessarily try to contact the discovery, since it's not even used for StaticTokenGetter.

**Version**
HEAD

PR #601, opened 2022-08-07
Fix issue #600. When using StaticTokenGetter do not make network calls.

vaikas:
When using StaticTokenGetter do not make network calls.

Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

#### Summary
Fix Issue #600 
When using StaticTokenGetter do not use OIDC discovery so that this flow works even without network connectivity.


hectorj2f (approved):
Lgtm

dlorenc (approved):

dekkagaijin (approved):

PR #602, opened 2022-08-08
Use well-known OIDC config for device endpoints

bobcallaway:
This patch adds the ability to use the well-known OIDC configuration to discover the device authorization and token endpoints used during a device code flow to fetch an identity token. This enables support for device code flow for OIDC providers other than `oauth2.sigstore.dev`.

Signed-off-by: Bob Callaway <bcallaway@google.com>

### DIFF
--- a/pkg/oauthflow/device.go
+++ b/pkg/oauthflow/device.go
@@ -204,7 +204,10 @@ func (d *DeviceFlowTokenGetter) CodeURL() (string, error) {
 
 	wellKnown := strings.TrimSuffix(d.Issuer, "/") + "/.well-known/openid-configuration"
 	/* #nosec */
-	resp, err := http.Get(wellKnown)
+	httpClient := &http.Client{
+		Timeout: 3 * time.Second,
+	}
+	resp, err := httpClient.Get(wellKnown)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


pkg/oauthflow/device.go
```diff
@@ -173,3 +195,46 @@ func (d *DeviceFlowTokenGetter) GetIDToken(p *oidc.Provider, cfg oauth2.Config)
...
 	}, nil
 }
+
+// CodeURL fetches the device authorization endpoint URL from the provider's well-known configuration endpoint
+func (d *DeviceFlowTokenGetter) CodeURL() (string, error) {
+	if d.codeURL != "" {
+		return d.codeURL, nil
+	}
+
+	wellKnown := strings.TrimSuffix(d.Issuer, "/") + "/.well-known/openid-configuration"
```
haydentherapper, line 205: Is this token authenticated at this point? Otherwise, it could be used as a DOS vector for making a request to an untrusted website.

pkg/oauthflow/device.go
```diff
@@ -173,3 +195,46 @@ func (d *DeviceFlowTokenGetter) GetIDToken(p *oidc.Provider, cfg oauth2.Config)
...
+	/* #nosec */
+	resp, err := http.Get(wellKnown)
```
haydentherapper, line 207: Should add a timeout

pkg/oauthflow/device.go
```diff
@@ -173,3 +195,46 @@ func (d *DeviceFlowTokenGetter) GetIDToken(p *oidc.Provider, cfg oauth2.Config)
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("unable to read response body: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	p := struct {
```
haydentherapper, line 222: nit: i know go style likes short variables, but it's hard to know what `p` refers to

pkg/oauthflow/device.go
bobcallaway: By this point, a lower level method within the oidc library has already fetched this information and verified that it matches the intended provider. It just does not persist all the information in a usable way, so we need to re-fetch it here.

haydentherapper (approved):

dlorenc (approved):

PR #615, opened 2022-08-11
wip: add initial TUF v2 client

asraa:

#### Summary

Starter implementation of a v2 TUF client. With configured TUFOptions to make it easier to use in external clients, and to implement the specification at https://docs.google.com/document/d/1QWBvpwYxOy9njAmd8vpizNQpPti9rd5ugVhji0r3T4c/edit#


asraa:
cc @kommendorkapten @joshuagl still working on this, expect some testing to be done by tomorrow for some initial client code.

there's a running list of TODOs in the client code D: I may need to separate AIs for them.

pkg/tuf_v2/repository/map.json
```diff
@@ -0,0 +1,13 @@
+{
+    "repositories": {
+      "public-good-instance": ["https://storage.googleapis.com/sigstore-tuf-root/"]
+    },
+    "mapping": [
+      {
+        "paths":        ["*"],
+        "repositories": ["Public-Good-Instance"],
```
vaikas, line 8: Is the naming capitalization different between L3 intentional? Looking at the tap4 spec, I'd expect this to be: "public-good-instance".

pkg/tuf_v2/repository/map.json
```diff
@@ -0,0 +1,13 @@
+{
+    "repositories": {
+      "public-good-instance": ["https://storage.googleapis.com/sigstore-tuf-root/"]
```
vaikas, line 3: Could we add another entry to see how this might look for say public and private Sigstore instance. I think that would also help drive my understanding on how the 'paths' below would be configured (as an example for testing cases where the local TUF remote is at http://tuf.tuf-system.svc in scaffolding) so that the right things would happen.
I'd be happy to start testing these out with the policy-controller btw if that would help validate some of these use cases. Naive guess is that "paths" section for a custom one would list the URLs of Fulcio, Rekor, CTLog, etc. but again, I think an example would help here.

vaikas (commented):
Thank you for driving this!

haydentherapper:
What’s the status of this PR?

asraa:
This should be closed in favor of the base TUF client in sigstore-go.

PR #623, opened 2022-08-14
Add a more friendly error in pubkey failure

### LINKED ISSUE #622: Fix unfriendly error in verify
**Description**

When doing the quickstart, I have mistakenly passed my private key instead of my public key and got a not so friendly error:
```
cosign verify --key=/Users/rkatz/cosign.key rpkatz/test:v1
Error: loading public key: pem to public key: asn1: structure error: tags don't match (16 vs {class:1 tag:27 length:34 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} publicKeyInfo @2
main.go:62: error during command execution: loading public key: pem to public key: asn1: structure error: tags don't match (16 vs {class:1 tag:27 length:34 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} publicKeyInfo @2
```

In privatekey there is already a verification of the type, so we can use the same approach in public key and get something more friendly like:

```
Error: loading public key: pem to public key: Not a valid public key. Are you trying to pass private key instead?
main.go:62: error during command execution: loading public key: pem to public key: Not a valid public key. Are you trying to pass private key instead?
```

rikatz:
#### Summary
Closes #622

When passing a wrong public key to the verify function, it will return an unfriendly error. Instead we can just let users know they are passing the wrong public key format

#### Release Note
Return a friendly error when an invalid key format is passed

#### Documentation
No change in docs

### DIFF
--- a/pkg/cryptoutils/publickey.go
+++ b/pkg/cryptoutils/publickey.go
@@ -52,7 +52,13 @@ func UnmarshalPEMToPublicKey(pemBytes []byte) (crypto.PublicKey, error) {
 	if derBytes == nil {
 		return nil, errors.New("PEM decoding failed")
 	}
-	return x509.ParsePKIXPublicKey(derBytes.Bytes)
+	switch derBytes.Type {
+	case string(PublicKeyPEMType):
+		return x509.ParsePKIXPublicKey(derBytes.Bytes)
+	default:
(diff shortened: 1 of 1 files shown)


dlorenc (approved):

cpanato (approved):
thanks!

PR #633, opened 2022-08-21
Fix: support certificate pem type

k4leung4:
Signed-off-by: Kenny Leung <kleung@chainguard.dev>

#### Summary

Change is needed to get fulcio test to pass. See https://github.com/sigstore/fulcio/runs/7935433305?check_suite_focus=true

My understanding is that CertificatePEMType is also valid. If not, tests in fulcio will need to be fixed.
https://github.com/sigstore/fulcio/blob/main/pkg/server/grpc_server_test.go#L383

### DIFF
--- a/pkg/cryptoutils/publickey.go
+++ b/pkg/cryptoutils/publickey.go
@@ -53,6 +53,8 @@ func UnmarshalPEMToPublicKey(pemBytes []byte) (crypto.PublicKey, error) {
 		return nil, errors.New("PEM decoding failed")
 	}
 	switch derBytes.Type {
+	case string(CertificatePEMType):
+		fallthrough
 	case string(PublicKeyPEMType):
 		return x509.ParsePKIXPublicKey(derBytes.Bytes)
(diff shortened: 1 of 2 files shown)


dekkagaijin:
Would be good to add a test case

k4leung4:
> Would be good to add a test case

added test

dekkagaijin (approved):

haydentherapper:
Thanks for making a fix! However, can we revert this? This will cause issues if you try to pass a PEM-encoded certificate, the PKIX parsing will fail.

The Fulcio test is what's wrong, it should be creating a PEM-encoded public key with `PublicKeyPEMType`.

haydentherapper:
Ah, I see why the Fulcio tests broke. A PR added PEM block type checking to return a better error message. Previously, the PEM block type didn't matter, so even though the Fulcio test was constructing invalid test input, the tests still passed.

So yea, the correct fix is just to fix the Fulcio test, and revert this.

PR #638, opened 2022-08-22
Add support for parsing PKCS#1 priv/pub keys, SEC1 priv keys

haydentherapper:
This adds support for parsing RSA public and private keys, and SEC 1 EC
private keys. This should reduce unexpected format errors for library
users.

#### Release Note

Added support for parsing PKCS#1 private and public keys, and SEC 1 private keys

### DIFF
--- a/pkg/cryptoutils/privatekey.go
+++ b/pkg/cryptoutils/privatekey.go
@@ -31,7 +31,11 @@ import (
 
 const (
 	// PrivateKeyPEMType is the string "PRIVATE KEY" to be used during PEM encoding and decoding
-	PrivateKeyPEMType                PEMType = "PRIVATE KEY"
+	PrivateKeyPEMType PEMType = "PRIVATE KEY"
+	// ECPrivateKeyPEMType is the string "EC PRIVATE KEY" used to parse SEC 1 EC private keys
+	ECPrivateKeyPEMType PEMType = "EC PRIVATE KEY"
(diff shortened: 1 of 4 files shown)


dekkagaijin:
"PRIVATE KEY" can begin blocks for arbitrary key types, yes? Is there an RFC which disambiguates them?

haydentherapper:
PKCS#8 encoding is typically associated with "PRIVATE KEY", which the code should already be doing for marshalling.

haydentherapper:
It doesn't look like I have the permission to retrigger the Action workflow, can a maintainer? I can't repro the test failure locally.

dlorenc (approved):

dlorenc:
> "PRIVATE KEY" can begin blocks for arbitrary key types, yes? Is there an RFC which disambiguates them?

There's no RFC here, it's very much just convention unfortunately. From this amazing blog post:
https://smallstep.com/blog/everything-pki/

> If you’re lucky [RFC 7468](https://tools.ietf.org/html/rfc7468) will give good guidance to figure out what your PEM payload is. Elliptic curve keys are usually labeled as such, though there [doesn’t seem to be any standardization](https://tools.ietf.org/html/rfc5915#section-4). Other keys are simply “PRIVATE KEY” by PEM. This usually indicates a [PKCS#8](https://tools.ietf.org/html/rfc5208) payload, an envelope for private keys that includes key type and other metadata. Here’s an example of a PEM-encoded elliptic curve key:

dekkagaijin:
Fair enough. Security libraries can have a little parsing ambiguity, as a treat :p

dekkagaijin (approved):

PR #677, opened 2022-09-09
Better define the project

lukehinds:
This should help clear any confusion new visitors may have over the projects purpose.

Also add missing CONTRIBUTORS.md 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,17 @@
-# sigstore framework
+# sigstore go library
 [![Fuzzing Status](https://oss-fuzz-build-logs.storage.googleapis.com/badges/sigstore.svg)](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=-opened&can=1&q=proj:sigstore) [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/5716/badge)](https://bestpractices.coreinfrastructure.org/projects/5716)
 
-sigstore/sigstore is a generic library / framework that is utilized by various other
-clients and projects including fulcio (webPKI), cosign (container and OCI signing tool)
-and tektoncd/chains (Supply Chain Security in Tekton Pipelines).
+sigstore/sigstore is a go library that provides API's for common sigstore
+features. If you need to develop a go based application that leverages
+sigstore, you're in the right place.
 
-sigstore is a good candidate for anyone wanting to develop go based clients / systems
-and utilise existing go modules for common sigstore functionality.
+sigstore/sigstore is utilized by various other clients and projects including 
+fulcio, rekor, cosign and others such as tektoncd/chains.
 
 This library currently provides:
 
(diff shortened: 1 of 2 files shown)


lukehinds:
cc @znewman01

JAORMX (approved):

asraa (approved):

README.md
```diff
@@ -1,17 +1,17 @@
-# sigstore framework
+# sigstore go library
 [![Fuzzing Status](https://oss-fuzz-build-logs.storage.googleapis.com/badges/sigstore.svg)](https://bugs.chromium.org/p/oss-fuzz/issues/list?sort=-opened&can=1&q=proj:sigstore) [![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/5716/badge)](https://bestpractices.coreinfrastructure.org/projects/5716)
 
-sigstore/sigstore is a generic library / framework that is utilized by various other
-clients and projects including fulcio (webPKI), cosign (container and OCI signing tool)
-and tektoncd/chains (Supply Chain Security in Tekton Pipelines).
+sigstore/sigstore is a go library that provides API's for common sigstore
+features. If you need to develop a go based application that leverages
+sigstore, you're in the right place.
```
znewman01, line 6: This suggests to me that it's *sufficient* to use Sigstore, which isn't true and is at odds with the claim below:

> sigstore/sigstore is utilized by various other clients and projects including 
fulcio [and] rekor[....]

`sigstore-python` and friends are much higher-level because they're not *used by* the Sigstore infra itself. If you wanted to do the same thing in Go, you'd have to cobble it together from `sigstore/sigstore`, `sigstore/rekor`, and `sigstore/fulcio`.

znewman01 (changes requested):
I have one complaint, but I feel pretty strongly about it 😄 

There's still ambiguity IMO in how this is worded.

README.md
lukehinds: @znewman01 I don't see the problem, its sufficient in its current scope (which is listed in the readme). Would you like me to state so somehow? what would be needed from sigstore and rekor?

znewman01: If I wanted to write a binary `myapp` that used Sigstore for something really basic, like the analog of `cosign sign-blob`, I don't think I could do it without a Rekor/Fulcio client. But for those I have to go to `sigstore/fulcio` and `sigstore/rekor`.  Even if I wrote those myself, I'd have to understand the flow of "generate key, then authenticate with Fulcio, then sign the artifact, then upload it to Rekor"—that's way too much detail IMO for an average client user. Something more like the API for [age](https://pkg.go.dev/filippo.io/age) is more appropriate. But we can't put that in s/s, because it would require a dependency on Fulcio and Rekor.

dlorenc:
Should this one be closed?

lukehinds:
> Should this one be closed?

Sure, will make another one but with only the contributors.md

PR #368, opened 2022-04-04
chore: set redirect URL in doOobFlow

hectorj2f:
#### Summary

I realized the RedirectURL was not reseted to `urn:ietf:wg:oauth:2.0:oob` even if a different value was set.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

PR #698, opened 2022-09-17
oauthflow/interactive: Fix oob flow

### LINKED ISSUE #594: Out of band flow not working
**Description**

The auth process in `gitsign` outputs a URL you should be able to open in the browser to get a verification code, but the oauth2 server redirects to localhost after logging in. In this case localhost is not accessible so the auth flow cannot complete.
```
$ git commit -m "foo"
error opening browser: exit status 3
Go to the following link in a browser:

         https://oauth2.sigstore.dev/auth/auth?access_type=online&client_id=sigstore&code_challenge=P69wZPuonUkCS2-LAp26XO-Ndw3GzeyATyuixsMRz7c&code_challenge_method=S256&nonce=xxx&redirect_uri=http%3A%2F%2Flocalhost%3A45077%2Fauth%2Fcallback&response_type=code&scope=openid+email&state=xxx
Enter verification code:
```

**Version**

gitsign v0.2.0

wlynch:
#### Summary

This change does 2 things:

1. Always sets oob URL for the oob flow. Reverts 8c6a840f - w.r.t. `pkg/oauth/oidc`: this looks largely duplicative of pkg/oauthflow so we probably want to delete / merge these together in another PR - both cosign and gitsign are using oauthflow. 🤷 
2. Changes Fscanln to Fscanf - Fscanln wasn't reading values unless it received EOF, making it non-obvious that a value was successfully read (especially in gitsign where we use the TTY directly so values don't populate to stdout).

Verified this works as expected for gitsign.

Fixes #594 

#### Release Note

🐛 Fixes OoB flow.

### DIFF
--- a/pkg/oauth/oidc/interactive.go
+++ b/pkg/oauth/oidc/interactive.go
@@ -37,9 +37,8 @@ const oobRedirectURI = "urn:ietf:wg:oauth:2.0:oob"
 type browserOpener func(url string) error
 
 func doOobFlow(cfg *oauth2.Config, stateToken string, opts []oauth2.AuthCodeOption) string {
-	if cfg.RedirectURL == "" {
-		cfg.RedirectURL = oobRedirectURI
-	}
+	cfg.RedirectURL = oobRedirectURI
+
 	authURL := cfg.AuthCodeURL(stateToken, opts...)
 	fmt.Fprintln(os.Stderr, "Go to the following link in a browser:\n\n\t", authURL)
 	fmt.Fprintf(os.Stderr, "Enter verification code: ")
--- a/pkg/oauthflow/interactive.go
+++ b/pkg/oauthflow/interactive.go
@@ -128,14 +128,15 @@ func (i *InteractiveIDTokenGetter) GetIDToken(p *oidc.Provider, cfg oauth2.Confi
 }
 
 func (i *InteractiveIDTokenGetter) doOobFlow(cfg *oauth2.Config, stateToken string, opts []oauth2.AuthCodeOption) string {
-	if cfg.RedirectURL == "" {
-		cfg.RedirectURL = oobRedirectURI
-	}
+	cfg.RedirectURL = oobRedirectURI
+
 	authURL := cfg.AuthCodeURL(stateToken, opts...)
 	fmt.Fprintln(i.GetOutput(), "Go to the following link in a browser:\n\n\t", authURL)
 	fmt.Fprintf(i.GetOutput(), "Enter verification code: ")
 	var code string
-	fmt.Fscanln(i.GetInput(), &code)
+	fmt.Fscanf(i.GetInput(), "%s", &code)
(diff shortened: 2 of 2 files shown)


wlynch:
@hectorj2f would appreciate a review here, since this is a revert of #368 (want to make sure this won't break something else as a result) 🙏

hectorj2f (approved):
Let me have a look

hectorj2f:
I remember when using [pinniped](https://github.com/vmware-tanzu/pinniped) as OIDC provider, we had to change the redirect url, that is why I initially created the PR.

hectorj2f:
@wlynch why do we always need `urn:ietf:wg:oauth:2.0:oob` ?

wlynch:
TBH I'm not 100% sure - I have to dig into the spec a bit more, but I do know that without this `oauth2.sigstore.dev` hangs and won't proceed to the code screen. 😭  See https://github.com/sigstore/sigstore/issues/594#issuecomment-1204580441

hectorj2f:
I am wondering if the oobRedirectURI only needs to be set by default when using `localhost` based on the comments. I don't want to block this PR, but It feels weird that we cannot use a specific redirect url is provided.

wlynch:
okay - I read the spec a bit, and I think this is correct mainly because we're using oob as a fallback if we fail to open the user's browser / can't use the localhost listener.

As is, redirectURL will never be empty here since this is [being set by the redirect listener before the call to doOOB](https://github.com/sigstore/sigstore/blob/181eb26ad2d04c11a387c35efdecef307c941541/pkg/oauthflow/interactive.go#L66). I don't think setting the redirect for the oob flow does much because [we only ever try to read the oob exchange code from stdin, we don't try to read it from the redirect listener](https://github.com/sigstore/sigstore/blob/181eb26ad2d04c11a387c35efdecef307c941541/pkg/oauthflow/interactive.go#L138). The oauth flow hangs because the server won't show the code since it's not `oob`, but the client is waiting for the user to copy+paste the code in.

We can support custom redirect URLs in the redirect listener via https://github.com/sigstore/sigstore/blob/181eb26ad2d04c11a387c35efdecef307c941541/pkg/oauthflow/interactive.go#L160
In most cases we should be able to use this with the browser based flow. For cases where we don't want to use the browser but still want to use the redirect listener, I think what we really want to do is make `browserOpener` configurable (e.g. make it echo, etc.). https://github.com/sigstore/sigstore/blob/181eb26ad2d04c11a387c35efdecef307c941541/pkg/oauthflow/interactive.go#L80 

For pinniped, what we may want to do is have a credential helper [provider](https://github.com/sigstore/cosign/tree/main/pkg/providers) that we can exec out to for a token. 

lmk if that makes sense! 🙏

hectorj2f (approved):

hectorj2f:
It sounds good to me.

ianlewis (approved):

PR #728, opened 2022-10-04
Handle invalid elliptic curve gracefully when verifying signature

haydentherapper:
An invalid curve caused a panic when verifying. Note that this does not affect Rekor or Fulcio. In both cases, an invalid curve will be caught when unmarshalling the PKIX public key, which checks for curve validity.

Caught by OSS-Fuzz!

### DIFF
--- a/pkg/signature/ecdsa.go
+++ b/pkg/signature/ecdsa.go
@@ -185,9 +185,9 @@ func (e ECDSAVerifier) VerifySignature(signature, message io.Reader, opts ...Ver
 		return fmt.Errorf("reading signature: %w", err)
 	}
 
-	// Without this check, VerifyASN1 panics on an invalid curve.
+	// Without this check, VerifyASN1 panics on an invalid key.
 	if !e.publicKey.Curve.IsOnCurve(e.publicKey.X, e.publicKey.Y) {
-		return fmt.Errorf("invalid ECDSA curve for %s", e.publicKey.Params().Name)
+		return fmt.Errorf("invalid ECDSA public key for %s", e.publicKey.Params().Name)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/signature/ecdsa.go
```diff
@@ -181,6 +181,11 @@ func (e ECDSAVerifier) VerifySignature(signature, message io.Reader, opts ...Ver
 		return fmt.Errorf("reading signature: %w", err)
 	}
 
+	// Without this check, VerifyASN1 panics on an invalid curve.
```
bobcallaway, line 184: is it worth a defensive check at the beginning of this function 

```
if e.publicKey == nil {
...
}
```

rare but you'd panic if that were the case

haydentherapper: Done

bobcallaway (dismissed):

pkg/signature/ecdsa.go
```diff
@@ -181,6 +185,11 @@ func (e ECDSAVerifier) VerifySignature(signature, message io.Reader, opts ...Ver
...
+	if !e.publicKey.Curve.IsOnCurve(e.publicKey.X, e.publicKey.Y) {
+		return fmt.Errorf("invalid ECDSA curve for %s", e.publicKey.Params().Name)
```
asraa, line 190: small nit: the ecdsa public key (point) is invalid (not on the given ECDSA curve)

asraa (dismissed):
nice tests!

pkg/signature/ecdsa.go
haydentherapper: Thanks! Updated.

asraa (approved):

dlorenc (approved):

PR #729, opened 2022-10-06
tuf: fix on-disk cache when writing targets in subfolders

asraa:

Added a test, patched `cosign` locally, and can confirm this now works correctly with the v5 root changes.

Note: no old clients will break from this change. It's forward compatible.

It's in anticipation of a root update that we'll push to a separate repository as to not break clients who are still using this version. Clients won't break until the current repository loses lifetime: in late January next year.

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -677,10 +677,12 @@ func (d *diskCache) Set(p string, b []byte) error {
 	if err := d.memory.Set(p, b); err != nil {
 		return err
 	}
-	if err := os.MkdirAll(d.base, 0o700); err != nil {
+
+	fp := filepath.FromSlash(filepath.Join(d.base, p))
+	if err := os.MkdirAll(filepath.Dir(fp), 0o700); err != nil {
 		return fmt.Errorf("creating targets dir: %w", err)
 	}
-	fp := filepath.FromSlash(filepath.Join(d.base, p))
+
 	return os.WriteFile(fp, b, 0o600)
 }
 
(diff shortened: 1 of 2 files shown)


haydentherapper:
To confirm, did you test the following after patching:
* Initialize to current (v4) metadata with no state
* Initialize to new (v5) metadata with no state
* Initialize to new metadata from state with current metadata

pkg/tuf/client.go
```diff
@@ -677,10 +677,12 @@ func (d *diskCache) Set(p string, b []byte) error {
 	if err := d.memory.Set(p, b); err != nil {
 		return err
 	}
-	if err := os.MkdirAll(d.base, 0o700); err != nil {
+
+	fp := filepath.FromSlash(filepath.Join(d.base, p))
+	if err := os.MkdirAll(filepath.Dir(fp), 0o700); err != nil {
```
haydentherapper, line 682: Silly question, what if the file exists but with different permissions?

haydentherapper (approved):

asraa:
Yes!

> To confirm, did you test the following after patching:

* Initialize to current (v4) metadata with no state: yep, that's `./cosign initialize`
* Initialize to new (v5) metadata with no state, yep, that's `./cosign initialize --mirror http://localhost:8001` immediately following the first 
* Initialize to new metadata from state with current metadata `rm` the dir and retry the previous

pkg/tuf/client.go
asraa: I think that's a valid cause of "your cache is corrupt" then, if you were manipulating it in a weird way.

haydentherapper:
Possibly related prober failure, investigating now - https://github.com/sigstore/sigstore/pull/729

haydentherapper:
Unrelated! Fixing issue

PR #734, opened 2022-10-10
Fix remoteFromMirror with GCS bucket

haydentherapper:
The mirror was reconfigured, but the URL was never reparsed, so when checking the scheme, there was a nil pointer exception. This fix simply returns early for GCS buckets.

#### Release Note

Fixed TUF root initialization with GCS bucket. This affects anyone who uses their own TUF root hosted on GCS, and specifies the GCS bucket only by name and not by HTTP path.


haydentherapper:
@cpanato We'll need to cut a new release of sigstore with this fix. If anyone is using their own GCS bucket for TUF in Cosign, they'll be affected.

bobcallaway (approved):

Issue #736, opened 2022-10-12
Tuf TestValidKey failing 


naveensrinivasan:
**Description**

The `TestValidKey` is failing with email being `0` and issuer `""`. The email `"test@rekor.dev` should be replaced with `0` and issuer `https://accounts.google.com` being replaced with `""`. 
https://github.com/sigstore/sigstore/blob/a9272f3e753d1d1fa374af7b5bba5263a5c06fbe/pkg/tuf/policy_test.go#L44-L65

Is this the expected behavior?

I was writing some fuzz tests for Tuf and ran into this issue.

naveensrinivasan:
Sounds good.

PR #737, opened 2022-10-12
Remove unused tuf code

naveensrinivasan:
#### Summary
- Removed tuf policy.go

#### Release Note
None


naveensrinivasan:
While doing this I ran into this #736

naveensrinivasan:
I would be adding additional Fuzz tests and then integrating with oss-fuzz

haydentherapper:
@asraa FYI

test/fuzz/tuf/fuzz_tuff_test.go
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+package tuf
+
+import (
+	"testing"
+
+	"github.com/sigstore/sigstore/pkg/tuf"
+)
+
+func FuzzTestAddKey(f *testing.F) {
```
haydentherapper, line 23: Will this test ever fail? AddKey adds a value to a map without verification of the structure of `publicKey`. I think this is testing a map insertion if so.

naveensrinivasan:
> Hi!
> 
> Note that this seems to be testing repository creation via `policy.go`: which is not used anywhere
> 
> It would be good to have a TUF client fuzzer, on the other hand, but that also interacts with a filesystem, not sure how easy that is to fuzz (the code in `client.go`)
> 
> Maybe this PR can actually remove the `policy.go` files entirely

I have removed the `policy.go` and its tests.

naveensrinivasan:
> > Hi!
> > Note that this seems to be testing repository creation via `policy.go`: which is not used anywhere
> > It would be good to have a TUF client fuzzer, on the other hand, but that also interacts with a filesystem, not sure how easy that is to fuzz (the code in `client.go`)
> > Maybe this PR can actually remove the `policy.go` files entirely
> 
> I have removed the `policy.go` and its tests.

@asraa   It is breaking the `signer.go`. Should that also be removed?

naveensrinivasan:
> Hi!
> 
> Note that this seems to be testing repository creation via `policy.go`: which is not used anywhere
> 
> It would be good to have a TUF client fuzzer, on the other hand, but that also interacts with a filesystem, not sure how easy that is to fuzz (the code in `client.go`)
> 
> Maybe this PR can actually remove the `policy.go` files entirely

Testing with `file system` is going to be an issue. Unless there is some abstraction AFAIK.

asraa:
> @asraa It is breaking the signer.go. Should that also be removed?

Yes!

asraa:
> Unless there is some abstraction AFAIK.

Enabling a test env var `SIGSTORE_NO_CACHE` disables the filesystem, but the problem is that the filesystem cache has had bugs before (and is the only major component off of go-tuf client itself)
We can configure the test dir to be a tempdir, and golang might work with an abstract fs impl

naveensrinivasan:
> > Unless there is some abstraction AFAIK.
> 
> Enabling a test env var `SIGSTORE_NO_CACHE` disables the filesystem, but the problem is that the filesystem cache has had bugs before (and is the only major component off of go-tuf client itself) We can configure the test dir to be a tempdir, and golang might work with an abstract fs impl

OK. Thanks

naveensrinivasan:
@asraa Friendly ping

rdimitrov:
Hey, does that mean that the following [cosign/cli/policy_init.go](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/policy_init.go) is unused and should be removed as well?

asraa:
> Hey, does that mean that the following [cosign/cli/policy_init.go](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/policy_init.go) is unused and should be removed as well?

Yes, I believe so - the only relevant TUF code is under the tuf package in this repo (which cosign pulls from) and has the client code!

asraa:
> pkg/tuf should provide anything TUF-related (client side) to the rest of the sigstore projects.

Correct!

> As a side note here, I se e some parts of TUF-related code that leaked out a bit into other projects/packages that should probably be moved to sigstore/pkg/tuf.

Hmm, I think there may be some accessors (like GetFulcio) that are leaked into other packages as intermediates. TBH, I think I agree that the tuf package should be all that's necessary.

That being said, those methods should be deprecated anyway and replaced with https://github.com/sigstore/sigstore/issues/1042

rdimitrov:
With the completion of https://github.com/sigstore/cosign/pull/2846, I think it should be safe to move forward with removing `policy.go` and `signer.go` too 👍

haydentherapper:
@rdimitrov did you want to take over removing those files? I’d say let’s just open a new pr

rdimitrov:
> @rdimitrov did you want to take over removing those files? I’d say let’s just open a new pr

Sure 👍 I'll open a PR at the end of next week(due to PTO).

PR #759, opened 2022-10-19
fulcioroots: Add support for SIGSTORE_ROOT_FILE.

wlynch:
#### Summary

There is a skew between s/s and cosign fulcioroots behavior, which leads to tools like gitsign not supporting the same environment variables. Instead of having each tool reimplement the same behavior, move this to s/s.

Signed-off-by: Billy Lynch <billy@chainguard.dev>

Part of https://github.com/sigstore/gitsign/issues/169

#### Release Note

fulcioroots package now respects the SIGSTORE_ROOTS_FILE environment variable.

### DIFF
--- a/pkg/fulcioroots/fulcioroots.go
+++ b/pkg/fulcioroots/fulcioroots.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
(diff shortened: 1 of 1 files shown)


wlynch:
cc @imjasonh (since you authored https://github.com/sigstore/cosign/pull/1866)

haydentherapper:
There was an intentional decision to leave out env variables from the core library to avoid unexpected behavior. Unless that's changed, I don't think we should add this.

haydentherapper:
cc @asraa

asraa:
I agree that sigstore/sigstore's fulcio root comes from TUF and the SIGSTORE_ROOT_FILE is a cosign implementation detail to configure custom verification. Cosign utilities also allow configuration of fulcio roots in the verification options.

SIGSTORE_ vars are named kind of broadly, partly because there weren't other clients before. SIGSTORE_ROOT_FILE is a super ambiguous name too that I would suggest you renaming to SOMETHING_FULCIO_ROOT_FILE anyway :)

It would consolidate some of the cosign logic though, so I'm not that opinionated about it; I would rather see better naming if it's in this top-level, or have env vars be project-specific.

haydentherapper:
At this point, given that we know we want to transform s/s, I'd just say to go with duplication in gitsign, and let's add this as an item to tackle within sigstore-go.

wlynch:
All for renaming! I just kept the existing name for compatibility.

w.r.t. to `SIGSTORE_` environment variables - I think it probably makes sense to have a set of known environment variables that works across tools - there are going to be common operations for Fulcio/Rekor like loading certs, configuring the URL, etc that aren't really specific to the tool itself it's probably going to be confusing if these differ slightly between tools. I think this is part of the reason why pulling in cosign as a dependency is so tempting, because it's the defacto standard for these config options.

I'm fine with deferring this to sigstore-go instead of s/s though! cc @znewman01

asraa:
> w.r.t. to SIGSTORE_ environment variables - I think it probably makes sense to have a set of known environment variables that works across tools

Absolutely +1, and to have other clients onboard if that happens. Let's start the work in sigstore-go though.

Issue #849, opened 2022-12-07
Azure KMS client does not handle HSM Key Vault keys


malancas:
**Description**

## Bug summary

Azure Key Vault supports creating EC and RSA key pairs in either the default storage or in validated HSMs. If the key pairs are stored in HSMs, the Key Type is suffixed with "-HSM". So "EC-HSM" or "RSA-HSM" are allowable key types.

Keys with the "HSM" suffixed key type are not handled by the `azureVaultClient#fetchPublicKey` method. When the [key is retrieved from Azure](https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/azure/client.go#L213), it has fields like `kty` and `kid`. `kty` represents the key type. This may be set to "EC", "EC-HSM", "RSA", or "RSA-HSM". When the method attempts to convert a key with a "HSM" suffixed key type to a [go-jose.JSONWebKey instance](https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/azure/client.go#LL224C34-L224C34), the operation fails with errors like `decoding the jsonWebKey: square/go-jose: unknown json web key type 'EC-HSM'`. 

This causes attempts to use the Azure KMS client with a key stored in the Key Vault HSMs to fail. As a solution, the `azureVaultClient#fetchPublicKey` method should be updated to handle incoming Azure Key Vault keys with key types suffixed with "-HSM".

## How to reproduce

Create a HSM backed key pair in Azure Key Vault and try to use the Azure KMS client with it, the client should fail to retrieve the key.

**Version**

 
N/A

malancas:
I have a pull request in progress to fix this, so I'll assign this issue to myself.

PR #850, opened 2022-12-07
Update Azure KMS client to handle HSM backed keys

malancas:
#### Summary

Closes https://github.com/sigstore/sigstore/issues/849

This adds support for handling HSM backed keys. Currently when the Azure KMS client tries to fetch and parse a key stored in Key Vault, it does not handle keys stored in HSMs. HSM backed keys have key types suffixed with "-HSM". This breaks when the client to tries to convert the key to a `go-jose.JSONWebKey` instance.

### DIFF
--- a/pkg/signature/kms/azure/client_test.go
+++ b/pkg/signature/kms/azure/client_test.go
@@ -16,9 +16,147 @@
 package azure
 
 import (
+	"context"
+	"crypto/elliptic"
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/base64"
+	"fmt"
 	"os"
+	"strings"
 	"testing"
+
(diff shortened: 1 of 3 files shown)


pkg/signature/kms/azure/client_test.go
```diff
@@ -16,10 +16,150 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/v7.1/keyvault"
 )
 
+type testKVClient struct {
+	key keyvault.JSONWebKey
+}
+
+func (c *testKVClient) CreateKey(ctx context.Context, vaultBaseURL string, keyName string, parameters keyvault.KeyCreateParameters) (result keyvault.KeyBundle, err error) {
+	key, err := generatePublicKey("EC")
```
haydentherapper, line 38: linter error

haydentherapper (approved):

PR #876, opened 2022-12-16
Improve the "Sigstore Auth Successful" page

### LINKED ISSUE #732: Improve the "Sigstore Auth Successful" page
Today, the post-OAuth page that gets displayed by tools like `cosign`, `gitsign` etc, is pretty ...sparse.

Screen Shot 2022-10-10 at 8 55 26 AM

That's not necessarily a bad thing. Being sparse also means it's not very resource intensive. It doesn't make any external requests to load CSS/JS/fonts/etc., and as the default for clients that may not have access to the public internet, that's a good thing.

At the same time, regular users of sigstore clients often end up with tons of these windows open (e.g., from `gitsign` and a long rebase chain, without the credential cache running).

For context, the source of that page is here:

https://github.com/sigstore/sigstore/blob/a862909dae84e62c1ac0eb356f311a73a62c0a1a/pkg/oauth/interactive.go#L18

## Closing the window 

We've learned (e.g., #484) that the page can't be closed using JavaScript. Windows can only reliably close windows that they themselves opened, and in our case the window was opened by `cosign` etc. We arrive at this page after some redirects through the OAuth flow.

This means that:
- the page can't autoclose after some wait
- the page can't include a button or click target that closes the window

If you try, you get `Scripts may close only the windows that were opened by them.`

To save you (and future-me) time, the StackOverflow results for that message include lots of obsolete hacks that browsers now block.

So closing the window is probably not feasible. Instead, we can at least make the page more _useful_.

## Making the page useful

New users are confronted with this page immediately after selecting their OAuth provider of choice. It's not really clear why you're here, what's happening, what to do next, etc. Close the window, sure, but then what?

We _don't_ have much context to make this page more specifically helpful to the user. The page URL is something like `http://localhost:59640/auth/callback?code=qdvscillhxxxxxxncrlbxhmtyg&state=2Fwdxxxx47dER2tw3I0ElvIgvxxx` -- there's (purposefully, I think) no information about what user just authed. At the point in the flow where this page is opened, Fulcio hasn't been contacted yet, so there's no information about the cert it returned. There's no Rekor entry yet to link to.

I think we can at least provide more text about what's happening, and where to go to learn more. A link to some page on docs.sigstore.dev describing the OAuth flow a bit better maybe, or a brief description of what just happened, what's happening now, etc.

The page contents are configurable by clients, so `cosign` could include text like "go back to the terminal where you invoked `cosign` to see more". Sigstore's own tools should at least aim to make these pages look and act consistently with each other.

## Making the page prettier

While (/if) we're adding content to read, we should also consider making the page easier to read. Nothing against black 12pt Times New Roman, but we've discovered over the last 20 years it may not be the best thing to read on a screen.

A few lines of CSS could go a long way to improving the experience.

## Considerations

The page shouldn't load any _external_ resources (e.g., JS/CSS/fonts) from the internet. Since `cosign`/`gitsign` are general-purpose clients, we can't guarantee clients have access to these resources at all. Even if they do, clients may not want to communicate to those external services that they're using these tools, and how or when.

This constraint should also help us avoid making the page _too_ flashy -- if we have to bundle everything into the page we're probably not going to make it some monstrous 15 MB dynamic-loading single-page app either.

---

This is mainly an issue to collect ideas and previous findings. If there's something you've ever wanted to see from the auth-success page, please tell us!

cc @wlynch @znewman01

julienv3:
#### Summary

This updates the "Sigstore Auth Successful" page to be visually consistent with https://sigstore.dev and include a few links.
Closes #732

#### Release Note

Added styling and documentation links to Sigstore Auth Successful page.

#### Documentation

No documentation update required.


julienv3:
image

https://user-images.githubusercontent.com/18708970/208181751-c397e7f8-51c3-431b-aa71-eaae12bb1922.mov

pkg/oauth/interactive.go
```diff
@@ -17,12 +17,59 @@ package oauth
...
+				</div>
+				<div class="content">
+					<span>You may now close this page.</span>
+				</div>
+			</div>
+			<div class="anchor">
+				<div class="links">
+					<a href="https://docs.sigstore.dev/" class="link login"><span class="sigstore">sigstore</span> documentation <span class="arrow">→</span></a>
+					<a href="https://blog.sigstore.dev/" class="link"><span class="sigstore">sigstore</span> blog <span class="arrow">→</span></a>
+					<a href="https://docs.sigstore.dev/gitsign/overview/" class="link login"><span class="sigstore">gitsign</span> documentation <span class="arrow">→</span></a>
```
haydentherapper, line 68: Can we drop this one, since this is documentation specific to just one client within Sigstore?

haydentherapper:
Please add a sign-off to your commit (`git commit -s`)

This looks really great!

pkg/oauth/interactive.go
julienv3: Removed third link and added sigstore home

haydentherapper:
Verified by compiling cosign with a locally modified sigstore library

haydentherapper (approved):

hectorj2f (approved):
lgtm

woodruffw:
This looks like really nice! We'll try to re-use it on `sigstore-python` as well 🙂

julienv3:
> Is it possible to avoid `script`? Would https://stackoverflow.com/a/59614656 work?

Unfortunately no, favicon can't reference a symbol in the page.
The JS is there to save space; the logo is a ~4k SVG, so I put it in the DOM and used JS to set the favicon using the same string.
We can very much hardcode the `<link>` to have the href that's currently dynamically computed avoiding script if that's preferrable?

haydentherapper:
I don't know enough about web dev to have a preference! I don't mind what's here currently, I wouldn't expect any issues with any browsers

haydentherapper:
Going to merge now, thanks for this improvement!

PR #893, opened 2023-01-04
Allow Initialize/NewFromEnv to reinitialize metadata if timestamp expired

haydentherapper:
This changes Initialize/NewFromEnv to always check if the timestamp is expired, while still locking access to the TUF object.

Ref: https://github.com/sigstore/public-good-instance/issues/1161

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -58,6 +58,9 @@ var (
 	singletonTUF     *TUF
 	singletonTUFOnce = new(sync.Once)
 	singletonTUFErr  error
+
+	// initMu locks concurrent calls to initializeTUF
+	initMu sync.Mutex
 )
 
 // getRemoteRoot is a var for testing.
@@ -240,6 +243,10 @@ func GetRootStatus(ctx context.Context) (*RootStatus, error) {
 // * forceUpdate: indicates checking the remote for an update, even when the local
 // timestamp.json is up to date.
 func initializeTUF(mirror string, root []byte, embedded fs.FS, forceUpdate bool) (*TUF, error) {
+	initMu.Lock()
+	defer initMu.Unlock()
+
+	var initErr error
 	singletonTUFOnce.Do(func() {
 		t := &TUF{
 			mirror:   mirror,
@@ -247,13 +254,13 @@ func initializeTUF(mirror string, root []byte, embedded fs.FS, forceUpdate bool)
 		}
 
 		t.targets = newFileImpl()
-		t.local, singletonTUFErr = newLocalStore()
-		if singletonTUFErr != nil {
+		t.local, initErr = newLocalStore()
+		if initErr != nil {
 			return
 		}
 
-		t.remote, singletonTUFErr = remoteFromMirror(t.mirror)
-		if singletonTUFErr != nil {
+		t.remote, initErr = remoteFromMirror(t.mirror)
+		if initErr != nil {
 			return
 		}
 
@@ -261,7 +268,7 @@ func initializeTUF(mirror string, root []byte, embedded fs.FS, forceUpdate bool)
 
 		trustedMeta, err := t.local.GetMeta()
 		if err != nil {
-			singletonTUFErr = fmt.Errorf("getting trusted meta: %w", err)
+			initErr = fmt.Errorf("getting trusted meta: %w", err)
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


haydentherapper:
A change in behavior to note: Previously, if you called Initialize, then NewFromEnv, the root would only be initialized once. Now, it will be twice. Additionally, Initialize is no longer thread safe with this implementation. 

Lemme know if you see a better way to do this. If you want Initialize to be thread safe, I'll use a mutex.

cpanato (approved):

asraa (changes requested):
/hold - we have had problems with thread safety with the TUF client

asraa (changes requested):
Could you move this https://github.com/sigstore/sigstore/blob/095df6c40e5a572fa6b6b9b7a2b83c1758549cf1/pkg/tuf/client.go#L280-L284 check outside the singleton, and if it fails, refresh the singleton?

Without the threadsafety running cosign in parallel will break, which has happened before

haydentherapper:
@asraa, I tried to move out all of the code before and including the timestamp expiration check outside of the singleton, but the concurrency test fails. I'm not certain what is causing this, so I thought it might be quicker to add a thread-unsafe method to use in the prober, which we need to unblock soon.

@vaikas, would you be able to take a look at this and let me know where there were thread safety issues in policy-controller?

vaikas:
The problems we were seeing surfaced there, but were fixed here (@asraa and I were collaborating on the fix, so there's two PRs but only one got merged):
https://github.com/sigstore/cosign/pull/1921
https://github.com/sigstore/cosign/pull/1932

There was also some other investigation here:
https://github.com/sigstore/cosign/pull/1941

Looks like the test for multiple access (TestConcurrentAccess) is there still in client_test.

There were also bunch of cleanup issues here:
https://github.com/sigstore/cosign/issues/1935

I jotted it down in the ☝️ but just stating it here as well, I'm curious about the need for this being implemented as a singleton instead of using RW locks while operating on the underlying data structures.

haydentherapper:
Thanks, I’ll take a look tomorrow!

Any opposition to what’s here? We are gonna have weekly alerts til we get a fix in, so I’d prefer to just get a quick fix in and refactor later.

pkg/tuf/client.go
```diff
@@ -313,10 +313,34 @@ func NewFromEnv(_ context.Context) (*TUF, error) {
 	}
 
 	// Initializes a new TUF object from the local cache or defaults.
-	return initializeTUF(mirror, nil, getEmbedded(), false)
+	return initializeTUFOnce(mirror, nil, getEmbedded(), false)
 }
 
+// Initialize creates a new TUF client, if one has not already been initialized,
+// and writes the cache to disk if requested. Thread safe.
```
vaikas, line 320: this doesn't seem threadsafe from a quick glance and from the PR description.

vaikas: Also probably worth while to note in the comments that this is the singleton instance. Or make it somehow clear that it does not necessairly return a new TUF client, but could return a previously created one. Reasoning being that it would then be easier to understand why this is different from InitializeAlways.

pkg/tuf/client.go
```diff
@@ -313,10 +313,34 @@ func NewFromEnv(_ context.Context) (*TUF, error) {
...
+			return err
+		}
+		if err := os.WriteFile(cachedRemote(rootCacheDir()), b, 0o600); err != nil {
+			return fmt.Errorf("storing remote: %w", err)
+		}
+	}
+	return nil
+}
+
+// InitializeAlways creates a new TUF client. Repeated calls to Initialize
```
vaikas, line 341: nit, Repeated calls to InitializeAlways?
This now allows one to have multiple clients simultaneously which I was having a bit of a rub before. I think by not forcing a singleton is nice, but the one thing that might be problematic still is the use of `rootCacheDir` below since now different clients will have a shared state in the form of the local cache, which is problematic. I think if you allow the caller to specify the cachedir here would be better.
And while you're at it, I think rather than having the getEmbedded live inside the function, taking in the fs.FS as an argument would be better, and would probably make the testing easier so one doesn't have to override the function (which IIRC caused some hilarity in the past as well. If not here, surely elsewhere :) ).

asraa: >  figure those should be hoisted to sigstore TUF client (something like GetFulcioRoots) since that's what (I think at least) the sigstore TUF client is adding on top of raw TUF, semantic meaning to various keys.

YES! I would have liked an API like:

````
GetTufObj(TufParams) -> TufObj

TufObj.GetFulcio -> Fulcio
```

asraa:
Idea maybe to explore: lock the singleton resets with a read/write mutex (that way it's safe to reset the singleton and block any readers from accessing the tufObj)

haydentherapper:
@vaikas @asraa I'm going to be taking another look at this, lemme know if you have any other thoughts. Otherwise I'll explore moving away from a singleton to mutexes.

haydentherapper:
@asraa @vaikas Reworked this, ptal

What I kept:
* A singleton TUF object that's initialized once per process

Changes:
* The singleton wraps the initialization of the TUF object, but not the checking of the timestamp (Asra, this was your suggestion earlier, idk why i didn't grok that)
* Every invocation of Initialize or NewFromEnv will check if the timestamp is expired and download updated metadata if expired
* A mutex now guards `initializeTUF`, which will prevent concurrent attempts to download updated metadata.

pkg/tuf/client.go
```diff
@@ -270,33 +277,39 @@ func initializeTUF(mirror string, root []byte, embedded fs.FS, forceUpdate bool)
...
+	if initErr != nil {
+		return nil, initErr
+	}
 
-		// Update if local is not populated or out of date.
-		if err := t.updateMetadataAndDownloadTargets(); err != nil {
-			singletonTUFErr = fmt.Errorf("updating local metadata and targets: %w", err)
-			return
-		}
+	trustedMeta, err := singletonTUF.local.GetMeta()
```
bobcallaway, line 296: On a subsequent call to this (if the first call failed for some reason), `singletonTUF` might still be nil and therefore panic here.

haydentherapper: Thanks, this should be fixed now, using `singletonTUFErr` like what was in the code before

vaikas (approved):

bobcallaway (approved):

asraa (commented):
Thanks! LGTM, sorry for the late review.

PR #962, opened 2023-02-10
update to use go1.20

cpanato:
#### Summary
- update to use go1.20


haydentherapper:
Looks like go 1.20.1 was just released, can you update to that too?

haydentherapper:
I’d be good with that, though I don’t recall if it gives us much since generics were in 1.18.

cpanato:
> I’d be good with that, though I don’t recall if it gives us much since generics were in 1.18.

will propose in a follow up pr

PR #969, opened 2023-02-13
Migrate from gopkg.in/square/go-jose.v2 to github.com/go-jose/go-jose/v3

mtrmac:
#### Summary
https://github.com/square/go-jose/tree/master says the former is deprecated. Moving everything to /v3 will, eventually, allow callers to only contain one vendored implementation instead of up to 3.

#### Release Note
NONE
#### Documentation
N/A


cpanato (approved):

cpanato:
thank you

Issue #1048, opened 2023-03-22
Panic `http: invalid pattern` when configuring a callback URL without path


mandelsoft:
**Description**

When configuring an empty path for the OIDC callback URL the used http MUX panics with "http: invalid pattern".

**Version**

v1.16

PR #1049, opened 2023-03-22
allow for empty callback URL path

mandelsoft:
#### Summary
When configuring an empty path for the OIDC callback URL the used http MUX panics with "http: invalid pattern".

This fix uses the MUX only for non-empty callback paths. If the path is empty the handler is directly used
for the local http server without any MUX.

#### Release Note

Support for OIDC callback URLs without path for usage of interactive mode

#### Documentation

Fixes: #1048

### DIFF
--- a/pkg/oauthflow/interactive.go
+++ b/pkg/oauthflow/interactive.go
@@ -191,16 +191,7 @@ func startRedirectListener(state, htmlPage, redirectURL string, doneCh chan stri
 		}
 	}
 
-	m := http.NewServeMux()
-	s := &http.Server{
-		Addr:    urlListener.Host,
-		Handler: m,
-
-		// an arbitrary reasonable value to fix gosec lint error
-		ReadHeaderTimeout: 2 * time.Second,
-	}
-
-	m.HandleFunc(urlListener.Path, func(w http.ResponseWriter, r *http.Request) {
+	var h http.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// even though these are fetched from the FormValue method,
 		// these are supplied as query parameters
 		if r.FormValue("state") != state {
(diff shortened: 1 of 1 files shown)


haydentherapper:
In what case is the callback URL empty?

mandelsoft:
I wanted to use with a local OIDC provider which was already configured with fixed callback URL for online scenario,

haydentherapper:
@bobcallaway What do you think about this? I’m not sure the use case, but it does clean up an error condition

bobcallaway:
> @bobcallaway What do you think about this? I’m not sure the use case, but it does clean up an error condition

I'd prefer something like this (mostly b/c it has a test, and it will only listen on `/` instead of any URL endpoint like the code in this PR):

```
diff --git a/pkg/oauthflow/interactive.go b/pkg/oauthflow/interactive.go
index dfc1f0c..eee7bad 100644
--- a/pkg/oauthflow/interactive.go
+++ b/pkg/oauthflow/interactive.go
@@ -184,6 +184,9 @@ func startRedirectListener(state, htmlPage, redirectURL string, doneCh chan stri
                if err != nil {
                        return nil, nil, err
                }
+               if urlListener.Path == "" {
+                       urlListener.Path = "/"
+               }
 
                listener, err = net.Listen("tcp", urlListener.Host)
                if err != nil {
diff --git a/pkg/oauthflow/interactive_test.go b/pkg/oauthflow/interactive_test.go
index 78591bc..3c55811 100644
--- a/pkg/oauthflow/interactive_test.go
+++ b/pkg/oauthflow/interactive_test.go
@@ -16,10 +16,27 @@ package oauthflow
 
 import (
        "bytes"
+       "net/http"
+       "net/http/httptest"
        "os"
        "testing"
 )
 
+// TestIssue1048 ensures that startRedirectListener does not panic due to the lack of a specified path in the redirectURL
+func TestIssue1048(t *testing.T) {
+       doneCh := make(chan string)
+       doneErr := make(chan error)
+       // this tests calling startRedirectListener with a redirectURL without a trailing slash
+       s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }))
+       t.Log(s.URL)
+       defer func() {
+               close(doneCh)
+               close(doneErr)
+               s.Close()
+       }()
+       startRedirectListener("something", "<html><body>something</body></html>", s.URL, doneCh, doneErr)
+}
+
 func TestInteractiveFlow_IO(t *testing.T) {
        t.Run("default", func(t *testing.T) {
                f := &InteractiveIDTokenGetter{}
```

haydentherapper:
@mandelsoft Would you like to update this PR to what Bob posted?

haydentherapper:
Closing due to no activity, feel free to reopen

PR #1054, opened 2023-03-24
Add additional EC algorithms supported by Azure KMS

malancas:
#### Summary

This adds additional EC algorithms supported by Azure's KMS offering Key Vault. See https://learn.microsoft.com/en-us/azure/key-vault/keys/about-keys-details for all supported keys and details.

I've tested this branch with a private repository using the added algorithms to confirm these updates fix the issue.

### DIFF
--- a/pkg/signature/kms/azure/client.go
+++ b/pkg/signature/kms/azure/client.go
@@ -303,9 +303,24 @@ func (a *azureVaultClient) createKey(ctx context.Context) (crypto.PublicKey, err
 	return a.public()
 }
 
-func (a *azureVaultClient) sign(ctx context.Context, hash []byte) ([]byte, error) {
+func getKeyVaultSignatureAlgo(algo crypto.Hash) keyvault.JSONWebKeySignatureAlgorithm {
+	switch algo {
+	case crypto.SHA256:
+		return keyvault.ES256
+	case crypto.SHA384:
+		return keyvault.ES384
+	case crypto.SHA512:
+		return keyvault.ES512
+	default:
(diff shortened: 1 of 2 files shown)


pkg/signature/kms/azure/client.go
```diff
@@ -303,9 +303,27 @@ func (a *azureVaultClient) createKey(ctx context.Context) (crypto.PublicKey, err
 	return a.public()
 }
 
-func (a *azureVaultClient) sign(ctx context.Context, hash []byte) ([]byte, error) {
+func getKeyVaultSignatureAlgo(algo crypto.Hash) (keyvault.JSONWebKeySignatureAlgorithm, error) {
```
malancas, line 306: The code does check when the SignerVerifier is created that it uses a supported algorithm, so this shouldn't ever return a non-nil error. I added it here for completion. I can add a test for this if we want to keep the ability to return an error.

pkg/signature/kms/azure/signer.go
```diff
@@ -152,7 +158,7 @@ func (a *SignerVerifier) VerifySignature(sig, message io.Reader, opts ...signatu
 		return fmt.Errorf("reading signature: %w", err)
 	}
 
-	// Convert the ANS.1 Sequence to a concantenated r||s byte string
+	// Convert the ANS.1 Sequence to a concatenated r||s byte string
```
haydentherapper, line 161: ASN

haydentherapper (approved):

haydentherapper (approved):

Issue #1063, opened 2023-03-31
Go version mismatch between go.mod and CI


clfs:
**Description**

Since PR https://github.com/sigstore/sigstore/pull/962, CI runs as Go 1.20: 

https://github.com/sigstore/sigstore/blob/589100556a55096e80e6b3c0ca6028da6fb9adbb/.github/workflows/e2e_test.yml#L25

It may make sense to upgrade the [go directive](https://go.dev/ref/mod#go-mod-file-go) in go.mod files to match:

https://github.com/sigstore/sigstore/blob/589100556a55096e80e6b3c0ca6028da6fb9adbb/go.mod#L3

PR #1064, opened 2023-03-31
Upgrade go.mod files to Go 1.20

clfs:
#### Summary

Addresses https://github.com/sigstore/sigstore/issues/1063:

>Since PR https://github.com/sigstore/sigstore/pull/962, CI runs as Go 1.20: 
>
>https://github.com/sigstore/sigstore/blob/589100556a55096e80e6b3c0ca6028da6fb9adbb/.github/workflows/e2e_test.yml#L25
>
>It may make sense to upgrade the [go directive](https://go.dev/ref/mod#go-mod-file-go) in go.mod files to match:
>
>https://github.com/sigstore/sigstore/blob/589100556a55096e80e6b3c0ca6028da6fb9adbb/go.mod#L3

Go 1.18 was released ~a year ago; this PR brings Sigstore onto the latest stable version of Go. 

I tested this PR against `make` targets. E2E tests, linting, etc. succeed.

#### Release Note

Upgraded module go directives to Go 1.20.

#### Documentation

I linked the wrong repo below; I haven't identified any documentation changes needed.

~I think this only requires one update:~

~- https://docs.sigstore.dev/cosign/installation/#with-go-119 should read `With Go 1.20`~

~I checked by searching "1.17", "1.18", "1.19", and "1.20" in the [sigstore/docs]( https://github.com/sigstore/docs) repo.~

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore
 
-go 1.18
+go 1.20
 
 require (
 	cloud.google.com/go/kms v1.10.0
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore
 
-go 1.20
+go 1.18
 
 require (
 	cloud.google.com/go/kms v1.10.0
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore/hack/tools
 
-go 1.17
+go 1.20
 
 require (
 	github.com/AdaLogics/go-fuzz-headers v0.0.0-20211123104302-8fea106b46e2
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore/hack/tools
 
-go 1.20
+go 1.18
 
 require (
 	github.com/AdaLogics/go-fuzz-headers v0.0.0-20211123104302-8fea106b46e2
--- a/test/fuzz/go.mod
+++ b/test/fuzz/go.mod
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore/test/fuzz
 
-go 1.18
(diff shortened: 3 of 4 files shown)


haydentherapper (changes requested):
Since we aren’t using any features of go 1.20, we had decided to keep it at 1.18

clfs:
@haydentherapper fair enough, thanks.

hack/tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore/hack/tools
 
-go 1.17
+go 1.20
```
cpanato, line 3: i think here we can bump to go1.20 with no problem

test/fuzz/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore/test/fuzz
 
-go 1.18
+go 1.20
```
cpanato, line 3: here as well

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore
 
-go 1.18
+go 1.20
```
cpanato, line 3: echoing @haydentherapper and this means the minimum required version to build this project

I would keep this as is for now and maybe we bump to go1.19 in a near future

cpanato (changes requested):

clfs:
I reverted the top-level `go.mod` file to Go 1.18 per @cpanato's feedback.

haydentherapper (commented):
I'd prefer we keep the same version of Go for the whole library, otherwise development gets tricky if you only have the older version

cpanato:
sgtm to me as well @haydentherapper 
@clfs can you sync all in the same? thanks!

clfs:
All `go.mod` files now sit on Go 1.18. Sorry for not checking beforehand; I could have made this smoother :)

haydentherapper (approved):
Sorry for not approving earlier, looks good, thanks!

PR #1108, opened 2023-04-20
map GCS HTTP endpoints to new CDN for sigstore public-good production TUF root

bobcallaway:
This PR silently changes the use of the sigstore public-good-instance production TUF root from calling GCS's HTTP endpoints to using the newly created CDN https://tuf-repo-cdn.sigstore.dev

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -44,12 +44,12 @@ import (
 const (
 	// DefaultRemoteRoot is the default remote TUF root location.
 	DefaultRemoteRoot = "https://tuf-repo-cdn.sigstore.dev"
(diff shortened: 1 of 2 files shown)


pkg/tuf/client.go
```diff
@@ -44,6 +44,12 @@ import (
 const (
 	// DefaultRemoteRoot is the default remote TUF root location.
 	DefaultRemoteRoot = "https://tuf-repo-cdn.sigstore.dev"
+	// DefaultRemoteGCSBucket is the name of the GCS bucket that holds sigstore's public good production TUF root
+	DefaultRemoteGCSBucket = "sigstore-tuf-root"
```
haydentherapper, line 48: Do you want to make these private constants to avoid library consumers using them?

haydentherapper (approved):

PR #1115, opened 2023-04-21
kms: split KMS providers into separate Go modules

imjasonh:
This creates separate Go modules for `pkg/signature/kms/{aws,azure,gcp,hashivault}`, which each have their own dependencies.

This means that the dependency tree of `sigstore/sigstore` (the root module) is _much_ smaller -- see `go.mod` in the root of the repo. Things that only want to depend on GCP's KMS implementation can pull in that dep without also pulling in deps for AWS, Azure, etc.

This adds some complexity to CI testing: **❗running `go test ./...` from the root of the repo doesn't test submodules❗**

This might also mean we should create more tags in the repo when we cut tags. For example, in addition to a future Git tag `v1.8.0` corresponding to the Go module `github.com/sigstore/sigstore@v1.8.0`, if we want these sub-Go-modules to be available as versioned Go modules, we must also add tags named like `pkg/signature/kms/aws/v1.8.0` (presumably tagged at the same commit as `v1.8.0`). See https://go.dev/ref/mod#vcs-dir for more info.

If we decide not to do that ☝️ , then `github.com/sigstore/sigstore/pkg/signature/kms/aws` will only be available by [pseudo-versions](https://go.dev/ref/mod#pseudo-versions) (e.g., `v0.0.0-20191109021931-daa7c04131f5`).


pkg/signature/kms/gcp/go.mod
```diff
@@ -0,0 +1,43 @@
+module github.com/sigstore/sigstore/pkg/signature/kms/gcp
+
+replace github.com/sigstore/sigstore => ../../../../
```
haydentherapper, line 3: I believe the replace directive breaks installation now, is it needed?

pkg/signature/kms/gcp/go.mod
```diff
@@ -0,0 +1,43 @@
...
+
+go 1.20
```
haydentherapper, line 5: Do we want to stick with 1.18?

haydentherapper (commented):
Can we leverage Go workspaces, https://go.dev/doc/tutorial/workspaces? I think this would simplify testing

haydentherapper:
> This might also mean we should create more tags in the repo when we cut tags. For example, in addition to a future Git tag v1.8.0 corresponding to the Go module github.com/sigstore/sigstore@v1.8.0, if we want these sub-Go-modules to be available as versioned Go modules, we must also add tags named like pkg/signature/kms/aws/v1.8.0 (presumably tagged at the same commit as v1.8.0). See https://go.dev/ref/mod#vcs-dir for more info.

This would break clients, right? Would we want to do a major release bump then?

haydentherapper:
+1 to releases of each submodule alongside the main module. just need to update (or add, I dont think we have any) release instructions

imjasonh:
> Can we leverage Go workspaces, https://go.dev/doc/tutorial/workspaces? I think this would simplify testing

Go workspaces could be useful too, yeah. Even if we go that route, we'd start by having separate Go modules in the repo, so this can be considered step 1 of getting there.

pkg/signature/kms/gcp/go.mod
imjasonh: I believe it can break `go install` _of binaries in this module_, but not those in subsequent downstream consumers of this module.

I might be wrong though, and we should try `go install`ing `cosign@main` et al depending on this commit to make sure, before we bump their deps.

imjasonh:
> This would break clients, right? Would we want to do a major release bump then?

Clients who depend on `sigstore/sigstore@1.6.3` who upgrade to `@1.6.4` would not be broken. When they `go get && go mod tidy` they would gain a new module dependency on whichever KMS provider modules they actually import, and lose all the unused transitive deps of the ones they don't.

imjasonh:
> +1 to releases of each submodule alongside the main module. just need to update (or add, I dont think we have any) release instructions

It'd probably be best to have some automation react to tagging `vX.Y.Z` by also tagging that as `pkg/signature/kms/$FOO/vX.Y.Z`.

imjasonh:
Friendly ping. I've made `make test` run tests for all modules, and `make tidy` run `go mod tidy` for all modules.

If you want I can write some GitHub automation to sync `<tag>` with `pkg/signature/kms/$foo/<tag>` as well.

haydentherapper:
Will take a look Monday, thanks for the ping

pkg/signature/kms/kms_test.go
```diff
@@ -1,35 +0,0 @@
...
-
-	"github.com/sigstore/sigstore/pkg/signature/kms"
-	"github.com/sigstore/sigstore/pkg/signature/kms/aws"
-	"github.com/sigstore/sigstore/pkg/signature/kms/azure"
-	"github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
-
-	"github.com/stretchr/testify/require"
-)
-
-func Test_SupportedProviders(t *testing.T) {
```
haydentherapper, line 29: Can we still have this test?

.github/workflows/sync-module-tags.yaml
```diff
@@ -0,0 +1,35 @@
...
+      matrix:
+        module:
+          - pkg/signature/kms/aws
+          - pkg/signature/kms/azure
+          - pkg/signature/kms/gcp
+          - pkg/signature/kms/hashivault
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          tag="${{ matrix.module }}/${{ github.ref_name }}"
```
haydentherapper, line 30: Just to check, it's `/`, not `@`, which will create the versioned module?

pkg/signature/kms/kms_test.go
imjasonh: This test would reintroduce all the newly split modules into dependencies of sigstore/sigstore. They would get trimmed out of built binaries since it's only a test dependency, but downstream `go.mod`s would still inherit these dependencies, and any source-level CVE warnings from GitHub for all transitive dependencies.

haydentherapper: Ah, that makes sense.

.github/workflows/sync-module-tags.yaml
imjasonh: That's my understanding, yes, based on [this](https://go.dev/ref/mod#vcs-version):

> If a module is defined in a subdirectory within the repository, that is, the [module subdirectory](https://go.dev/ref/mod#glos-module-subdirectory) portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash. For example, the module golang.org/x/tools/gopls is defined in the gopls subdirectory of the repository with root path golang.org/x/tools. The version v0.4.0 of that module must have the tag named gopls/v0.4.0 in that repository.

The [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go) repo is another example:

- the [root go.mod](https://github.com/open-telemetry/opentelemetry-go/blob/main/go.mod) corresponds to go.opentelemetry.io/otel (they use a vanity importpath, to confuse things)
- there's also [`metric/go.mod`](https://github.com/open-telemetry/opentelemetry-go/blob/main/metric/go.mod), with a tag [`metric/v0.38.1`](https://github.com/open-telemetry/opentelemetry-go/releases/tag/metric/v0.38.1) which makes that [module available at v0.38.1](https://pkg.go.dev/go.opentelemetry.io/otel/metric@v0.38.1)

imjasonh:
> Just to check, any updates needed for the dependabot script?

I believe dependabot is already aware of repos containing multiple Go modules, and will send separate PRs for dependency updates for sub-Go-modules. For example, in otel: https://github.com/open-telemetry/opentelemetry-go/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+is%3Aclosed

.github/workflows/sync-module-tags.yaml
haydentherapper: Thanks for checking!

haydentherapper:
Looks great, just need to rebase.

haydentherapper (approved):

PR #1141, opened 2023-05-04
Add comment to GCP KMS about ignored hash function

codysoyland:


haydentherapper:
Can you drop the unrelated commit?

haydentherapper:
Actually, would it be better to include this in Fulcio than here?

haydentherapper:
I was thinking of a comment in Fulcio, since here it ignored the parameter

PR #1149, opened 2023-05-04
Azure KMS: Infer hash function from key

codysoyland:
Azure KMS explicitly supports a [single hash function per EC curve](https://learn.microsoft.com/en-us/azure/key-vault/keys/about-keys-details#signverify). This PR makes the Azure KMS provider infer the hash function from the key. This brings the behavior closer to that of the GCP KMS provider, and allows Fulcio to use Azure KMS with curves longer than P-256. This is an alternative to https://github.com/sigstore/fulcio/pull/1152

### DIFF
--- a/pkg/signature/kms/azure/client.go
+++ b/pkg/signature/kms/azure/client.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rsa"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -43,8 +45,8 @@ import (
 )
 
 func init() {
-	sigkms.AddProvider(ReferenceScheme, func(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts ...signature.RPCOption) (sigkms.SignerVerifier, error) {
-		return LoadSignerVerifier(ctx, keyResourceID, hashFunc)
+	sigkms.AddProvider(ReferenceScheme, func(ctx context.Context, keyResourceID string, _ crypto.Hash, opts ...signature.RPCOption) (sigkms.SignerVerifier, error) {
+		return LoadSignerVerifier(ctx, keyResourceID)
 	})
 }
 
@@ -248,14 +250,7 @@ func (a *azureVaultClient) fetchPublicKey(ctx context.Context) (crypto.PublicKey
 		return nil, fmt.Errorf("decoding the jsonWebKey: %w", err)
 	}
 
-	pub, ok := jwk.Key.(*ecdsa.PublicKey)
-	if !ok {
-		if err != nil {
-			return nil, fmt.Errorf("public key was not ECDSA: %#v", pub)
-		}
-	}
-
-	return pub, nil
+	return jwk.Key, nil
 }
 
 func (a *azureVaultClient) getKey(ctx context.Context) (azkeys.KeyBundle, error) {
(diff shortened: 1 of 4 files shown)


pkg/signature/kms/azure/client.go
```diff
@@ -304,21 +304,47 @@ func (a *azureVaultClient) createKey(ctx context.Context) (crypto.PublicKey, err
 	return a.public(ctx)
 }
 
-func getKeyVaultSignatureAlgo(algo crypto.Hash) (keyvault.JSONWebKeySignatureAlgorithm, error) {
-	switch algo {
+func (a *azureVaultClient) getKeyVaultHashFunc() (crypto.Hash, error) {
```
haydentherapper, line 307: Maybe outside of scope for this PR, but should this client support more than just ECDSA keys? Let's at least document this somewhere (fwiw every client is a little different in this regard)

pkg/signature/kms/azure/client.go
```diff
@@ -304,21 +304,47 @@ func (a *azureVaultClient) createKey(ctx context.Context) (crypto.PublicKey, err
...
+	publicKey, err := a.public(context.TODO())
+	if err != nil {
+		return 0, fmt.Errorf("failed to get public key: %w", err)
+	}
+	switch eckey := publicKey.(type) {
+	case *ecdsa.PublicKey:
+		switch eckey.Params().Name {
```
haydentherapper, line 314: nit, I'd compare against `eckey.Curve`, each case statement checking if it's `elliptic.P256()` or 384 or 521.

pkg/signature/kms/azure/client.go
```diff
@@ -304,21 +304,47 @@ func (a *azureVaultClient) createKey(ctx context.Context) (crypto.PublicKey, err
 	return a.public(ctx)
 }
 
-func getKeyVaultSignatureAlgo(algo crypto.Hash) (keyvault.JSONWebKeySignatureAlgorithm, error) {
-	switch algo {
+func (a *azureVaultClient) getKeyVaultHashFunc() (crypto.Hash, error) {
+	publicKey, err := a.public(context.TODO())
```
malancas, line 308: Should we pass the `context` object passed as an argument to the `sign` method down to this `public` call?

pkg/signature/kms/azure/client.go
haydentherapper: GCP has the same for KMS keys, it looks like there's a hardcoded mapping between key name and hash algorithm - https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/gcp/client.go#L210-L232

Do we want to do the same here? Another benefit is it makes it clear which key types we support.

haydentherapper: So I started digging into this more. `SignMessage` is just a sigstore/sigstore interface, and not one called from `crypto/x509` `x509.CreateCertificate`, which calls `Sign` with a given digest. It looks like for RSA, it's hardcoded for sha256 - https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/crypto/x509/x509.go;l=1410

I was surprised to see this. I might be wrong, I haven't dug into this much, but if so, then it's relevant if you provide a different hash function for RSA.

haydentherapper: Sorry, completely ignore what I wrote above, that's just the default.

But we do still need to make a change in Fulcio to specify the SignatureAlgorithm.

haydentherapper: https://github.com/sigstore/fulcio/issues/1161 

We should probably expose the signature algorithm from sigstore/sigstore too.

codysoyland:
I added support for RSA and rebased everything. I think this is ready for review!

haydentherapper (approved):
Looks great! Thanks for adding support for both.

PR #1216, opened 2023-06-13
update golangci-lint to v1.53.x and add depguard rules

cpanato:
#### Summary
- update golangci-lint to v1.53.x and add depguard rules

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -55,4 +55,4 @@ jobs:
         uses: golangci/golangci-lint-action@639cd343e1d3b897ff35927a75193d57cfcba299 # v3.6.0
         timeout-minutes: 5
         with:
-          version: v1.52
+          version: v1.53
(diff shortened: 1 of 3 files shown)


haydentherapper (commented):
What is the purpose of depguard?

cpanato:
> What is the purpose of depguard?

> A Go linter that checks package imports are in a list of acceptable packages. This allows you to allow imports from a whole organization or only allow specific packages within a repository.


I was thinking and I am not sure if we want to have that. What are your thoughts?

haydentherapper:
I think this would make sense more at the org level rather than repo. Because if I want to add a dependency, I’ll just update this list in the same PR. 

I’m hesitant to put this at the org level though since that sounds like a lot of effort to maintain.

cpanato:
> I think this would make sense more at the org level rather than repo. Because if I want to add a dependency, I’ll just update this list in the same PR.
> 
> I’m hesitant to put this at the org level, thought since that sounds like a lot of effort to maintain.

agree and thinking more for now that is just overhead to us, i will drop that change and remove the depguard check, thanks for the review and feedback

haydentherapper (approved):

PR #1247, opened 2023-06-29
Specify hashes when comparing target meta.

kommendorkapten:
#### Summary

Before the default (sha512) hash function was the only one used when comparing hashes. This commit makes the choice explicit instead of depending on go-tuf's default vaule. Now sha256 and sha512 are recognized as valid hash functions.
Also this commit does not hide the error from go-tuf's underlying methods used to verify the target's meta.

#### Release Note
SHA256 is now a recognized hash function when verifying target metadata from a TUF repository

#### Documentation
N/A

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -370,15 +370,17 @@ func Initialize(_ context.Context, mirror string, root []byte) error {
 }
 
 // Checks if the testTarget matches the valid target file metadata.
-func isValidTarget(testTarget []byte, validMeta data.TargetFileMeta) bool {
-	localMeta, err := util.GenerateTargetFileMeta(bytes.NewReader(testTarget))
+func isValidTarget(testTarget []byte, validMeta data.TargetFileMeta) (bool, error) {
+	localMeta, err := util.GenerateTargetFileMeta(
+		bytes.NewReader(testTarget),
+		"sha256", "sha512")
 	if err != nil {
-		return false
+		return false, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/tuf/client.go
```diff
@@ -540,7 +542,7 @@ func maybeDownloadRemoteTarget(name string, meta data.TargetFileMeta, t *TUF) er
 	// If we already have the target locally, don't bother downloading from remote storage.
 	if cachedTarget, err := t.targets.Get(name); err == nil {
 		// If the target we have stored matches the meta, use that.
-		if isValidTarget(cachedTarget, meta) {
+		if valid, _ := isValidTarget(cachedTarget, meta); valid {
```
haydentherapper, line 545: do we need to check err here and below?

kommendorkapten: I don't think it's needed. This is the `maybeDownloadRemoteTarget`, so it's basically checks if the target exists locally, and that the metadata matches the one from upstream repo. So if the target is not valid, it may be due to a stale cache, so the client will just attempt to download the (updated) target from remote. So I don't think the actual error message is important here, as a stale cache would not necessarily indicate an issue. If there is a "real" issue, such that unsupported hash algorithm or so, the error will bubble up during the next `GetTarget` invocation.

bobcallaway:
@kommendorkapten https://github.com/sigstore/sigstore/pull/1263 should resolve the unit test issue

kommendorkapten:
Thanks @bobcallaway 
I'll rebase this on main once #1263 is merged

haydentherapper (approved):

PR #1170, opened 2023-05-18
Update Azure Key Vault client

malancas:
#### Summary

Update the underlying Azure SDK Key Vault client used by the Azure KMS client. This allows us to use newer functions and client defaults, include request retry defaults used when connecting with the Key Vault.

PR #1272, opened 2023-07-12
Add support for additional Azure clouds (bug fix),  add optional integration test

malancas:
#### Summary

This pull request address bugs reported in https://github.com/sigstore/cosign/issues/3094 and https://github.com/sigstore/cosign/issues/3095. It adds supports for the US Government cloud and China cloud. Previously the client would always default to using the public cloud. The client now looks for the `AZURE_ENVIRONMENT` environment variable to set the cloud. The underlying Azure SDK used by the client before https://github.com/sigstore/sigstore/pull/1170 was merged used this environment variable. Dropping support for this environment variable may be part of the cause for the bugs above.

This pull request also adds optional integration tests. This file is meant to be run locally by Azure client developers using an Azure subscription of their choice. Because Sigstore does not use Azure for its infrastructure, this file is not run in any CI/CD.


pkg/signature/kms/azure/client_test.go
```diff
@@ -235,3 +235,58 @@ func testSetEnv(t *testing.T, s map[string]string) func() {
 		}
 	}
 }
+
+func TestParseReference(t *testing.T) {
```
malancas, line 239: The content from `azure_test.go` was copied here. Since all of the tests in that file were around testing the client, it made sense to move them here.

pkg/signature/kms/azure/integration_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}, {
+		env:            "AZUREPUBLICCLOUD",
+		expectedConfig: cloud.AzurePublic,
+	}, {
+		env:            "",
+		expectedConfig: cloud.AzurePublic,
+	}}
+
+	for _, tc := range testCases {
+		err := os.Setenv("AZURE_ENVIRONMENT", tc.env)
```
bobcallaway, line 72: does this work with `t.Setenv` instead of `os.Setenv`? using `t.Setenv` will ensure it gets unset after the test case is run.

malancas: Yes, just tested that this works with `t.Setenv`. I'll push an update.

PR #1312, opened 2023-08-07
Bump TUF root version

haydentherapper:
Also update the embedded targets.


haydentherapper:
@asraa, I wanted to confirm, these target file names should not be prefixed with their hash, correct? I looked through the code and believe that's accurate, but wanted to double check.

haydentherapper:
fyi @jku, since you had posted an issue about updating tuf metadata

asraa:
> @asraa, I wanted to confirm, these target file names should not be prefixed with their hash, correct? I looked through the code and believe that's accurate, but wanted to double check.

I'm pretty sure these are used to prepopulate the cache, and when you download to a local store TUF clients strip the hash prefix. I would test it out locally too.

znewman01 (approved):
I would love instructions for doing this change (and for verifying it) to be documented somewhere. Also maybe how frequently we should be doing it and maybe even issues that get automatically filed.

None of that's blocking for now!

haydentherapper:
https://github.com/sigstore/root-signing/issues/817 for tracking automation (across all client codebases too)

To verify, in Cosign, I bumped go-tuf to 0.6.0 and replaced s/s with my local repo. Verified without the replace directive `cosign initialize` fails, with the replace directive `cosign initialize` succeeds.

I've also removed `set_ecdsa` from s/s because we no longer need to support the deprecated verifier.

haydentherapper:
Removing `set_ecdsa` is technically a breaking change, if a client hasn't requested a new root since the generation of v5, since the TUF client will have to load the v2/v3/v4 root keys to verify subsequent roots. That was 10 months ago though. I would say that this change is fine to do.

znewman01 (approved):
nice!

haydentherapper:
Fixing tests now, this is from removing set_ecdsa...

znewman01 (approved):

bobcallaway (approved):

PR #1414, opened 2023-09-22
Handles the error in kms/hashivault

PuneetPunamiya:
#### Summary

* While using Tekton Chains with hashivault, if there was some misconfiguration, then it was getting panic because the error was not handled correctly

* Hence this patch handles the error before returning the value


haydentherapper (approved):
Thanks!

PR #1426, opened 2023-10-03
Add a hash algorithm method to the KMS clients

malancas:
#### Summary

Opened as part of a discussion on https://github.com/sigstore/timestamp-authority/pull/488, this adds a `HashFunc` method (open to any name suggestions) to the KMS clients. This method will return the hash algorithm used with the signing key. I opened this so we could use the signing key has function in sigstore/timestamp-authority without having to pass the hash function as a flag to the server. See [this PR](https://github.com/sigstore/timestamp-authority/pull/517) for how the new KMS method is used in the TSA.


pkg/signature/kms/kms.go
```diff
@@ -72,7 +72,14 @@ func SupportedProviders() []string {
 type SignerVerifier interface {
 	signature.SignerVerifier
 	CreateKey(ctx context.Context, algorithm string) (crypto.PublicKey, error)
-	CryptoSigner(ctx context.Context, errFunc func(error)) (crypto.Signer, crypto.SignerOpts, error)
+	CryptoSigner(ctx context.Context, errFunc func(error)) (CryptoSignerWrapper, crypto.SignerOpts, error)
```
haydentherapper, line 75: What's the motivation for this change over just exposing the additional `HashFunc` function? This would be a breaking change if anyone is expecting a crypto.Signer, correct?

Hayden-IO (commented):
LGTM overall, just a question about the interface change

Hayden-IO (approved):

haydentherapper:
The failure might be due to a aws mock that needs to be updated since there are different calls being made, though i haven't looked closely.

malancas:
> The failure might be due to a aws mock that needs to be updated since there are different calls being made, though i haven't looked closely.

Agreed, I'm going to take closer look at the aws mock.

PR #1428, opened 2023-10-05
use the latest key version

developer-guy:
#### Summary

based on the discussions [here](https://github.com/sigstore/cosign/issues/1351), we added a code to use the latest version of the rotated key.

_P.S. If you think this is wrong please feel free to close the PR_

### DIFF
--- a/pkg/signature/kms/hashivault/client.go
+++ b/pkg/signature/kms/hashivault/client.go
@@ -313,7 +313,27 @@ func (h hashivaultClient) verify(sig, digest []byte, alg crypto.Hash, opts ...si
 			if h.keyVersion > 0 {
 				vaultDataPrefix = fmt.Sprintf("vault:v%d:", h.keyVersion)
 			} else {
-				vaultDataPrefix = vaultV1DataPrefix
+				// use hashivault client to grab the latest version of the key and use that as the prefix
+				/*				vaultDataPrefix = vaultV1DataPrefix*/
+
+				client := h.client.Logical()
+
+				path := fmt.Sprintf("/%s/keys/%s", h.transitSecretEnginePath, h.keyPath)
+
+				keyResult, err := client.Read(path)
+				if err != nil {
+					return fmt.Errorf("public key: %w", err)
+				}
+
+				if keyResult == nil {
+					return fmt.Errorf("could not read data from transit key path: %s", path)
+				}
+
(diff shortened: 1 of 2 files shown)


trishankatdatadog:
I've [programmed](https://github.com/theupdateframework/python-tuf/blob/495e10f346342892b3ec645b3ce488ede64dd30d/tuf/api/keys.py) a similar integration in Vault before, and wonder whether it's a good idea to hit the Vault API just to verify a signature 🙂 Because you really only need (assuming keys are not exportable) the Vault Transit Secrets Engine to make a signature but not verify it. This would yield in better performance and reliability, but you do need knowledge about the public key from somewhere else. WDYT?

Hayden-IO (commented):
I'll let Bob review this PR since he has more context.

I'm mostly familiar with GCP KMS, which I believe allows either latest or a specific key version. Should we do the same here? Some may prefer a specific key version to avoid silently switching out the key.

trishankatdatadog:
> I'm mostly familiar with GCP KMS, which I believe allows either latest or a specific key version. Should we do the same here? Some may prefer a specific key version to avoid silently switching out the key.

There is certainly this issue also, yes. Perhaps default to latest is version is unspecified, but fetch the exact version when it is specified.

pkg/signature/kms/hashivault/client.go
```diff
@@ -313,7 +313,27 @@ func (h hashivaultClient) verify(sig, digest []byte, alg crypto.Hash, opts ...si
 			if h.keyVersion > 0 {
 				vaultDataPrefix = fmt.Sprintf("vault:v%d:", h.keyVersion)
 			} else {
-				vaultDataPrefix = vaultV1DataPrefix
+				// use hashivault client to grab the latest version of the key and use that as the prefix
```
bobcallaway, line 316: I have one concern with this implementation - it ignores the `keyCache` completely; I'd rather have it invalidate the cache, call `public()` (which would hydrate the cache with the `latest_version`. 

It still results in another round trip to Vault to do the verification, but we at least have the key material for re-use in memory.

nia-potato:
Hi all, any updates on this issue? stuck with the same problem.

kent007:
Also interested in updates, has anyone looked at this recently?

haydentherapper:
Moving to draft. If someone would like to pick up this implementation, feel free to.

PR #1497, opened 2023-11-08
Fixes the error handling in kms/hashivault

PuneetPunamiya:
* Previously with this patch https://github.com/sigstore/sigstore/pull/1414, 
the error was handled but the order of the error was not correct, because
while using Tekton Chains with hashivault, chains controller was getting 
paniced because the order of error handled was not correct

* Hence, with this patch, error is handled correctly and also a condition is 
added to check if the item retrieved is nil or not to avoid Tekton Chains 
controller pod crashing

### DIFF
--- a/pkg/signature/kms/hashivault/client.go
+++ b/pkg/signature/kms/hashivault/client.go
@@ -246,12 +246,17 @@ func (h *hashivaultClient) public() (crypto.PublicKey, error) {
 			return nil
 		},
 	)
+
+	item := h.keyCache.Get(cacheKey, ttlcache.WithLoader[string, crypto.PublicKey](loader))
 	if lerr != nil {
 		return nil, lerr
 	}
 
-	item := h.keyCache.Get(cacheKey, ttlcache.WithLoader[string, crypto.PublicKey](loader))
-	return item.Value(), lerr
+	if item == nil {
+		return nil, fmt.Errorf("unable to retrieve an item from the cache by the provided key")
+	}
+
+	return item.Value(), nil
 }
 
 func (h hashivaultClient) sign(digest []byte, alg crypto.Hash, opts ...signature.SignOption) ([]byte, error) {


hectorj2f (approved):
lgtm

pkg/signature/kms/hashivault/client.go
```diff
@@ -246,12 +246,17 @@ func (h *hashivaultClient) public() (crypto.PublicKey, error) {
 			return nil
 		},
 	)
+
+	item := h.keyCache.Get(cacheKey, ttlcache.WithLoader[string, crypto.PublicKey](loader))
 	if lerr != nil {
```
cpanato, line 251: Maybe this should go up

cpanato (changes requested):

pkg/signature/kms/hashivault/client.go
PuneetPunamiya: Yeah before we added this up, but the chains controller pod was failing
Also as far as I understood, loader was called when we are assigning the value to `item` and hence if we add this error check up, at that time error is nil 
Adding it here, I think handles the error properly

haydentherapper: Good catch - If I'm understanding this correctly, lerr will always be nil in the previous code because it only gets set once the loader is invoked, which is in the `Get` call.

pkg/signature/kms/hashivault/client.go
```diff
@@ -246,12 +246,17 @@ func (h *hashivaultClient) public() (crypto.PublicKey, error) {
...
 		return nil, lerr
 	}
 
-	item := h.keyCache.Get(cacheKey, ttlcache.WithLoader[string, crypto.PublicKey](loader))
-	return item.Value(), lerr
+	if item == nil {
```
haydentherapper, line 255: Is this just to be defensive, or is there a case where an item won't be found in the cache and lerr will be nil?

pkg/signature/kms/hashivault/client.go
PuneetPunamiya: yes exactly, and hence even if there was an error, it was not returned earlier and the `item`  would be nil which will indeed panic the code
Hence with the current patch, the error would be handled correctly

pkg/signature/kms/hashivault/client.go
PuneetPunamiya: I would say, just to be more defensive, 
Because if we look at this [function](https://github.com/jellydator/ttlcache/blob/v3/cache.go#L205-L240) I see close to zero chances an item will be nil, if lerr is nil

haydentherapper (approved):

haydentherapper:
@cpanato Can you take a look too?

cpanato (approved):

cpanato:
thanks!

Issue #1175, opened 2023-05-21
AWS KMS fails to use STS when endpoint is provided


ChevronTango:
**Description**

When a user specifies an endpoint as part of their KMS config eg `awskms://[endpoint]/[arn]` then this only works if the client does not also have to call STS as well, such as when making a "AssumeRoleWithWebIdentity` call used by clients within a kubernetes cluster. STS has a different endpoint to KMS but https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/aws/client.go specifies a single endpoint override for all calls, meaning STS fails with an unknown command response as it tries to send the STS action to the KMS endpoint the user specified in their config string. If STS is being used, its endpoint shouldn't be overwritten by the same mechanism that overwrites the endpoint for KMS.

**Version**

fulcio helm

PR #1509, opened 2023-11-10
Add conditional endpoint resolver for AWS KMS

euanhume:
#### Summary

This PR introduces a conditional check within the endpoint resolver to specifically target the AWS Key Management Service (KMS). Currently, when using Security Token Service (STS), an 'unknown command' error occurred due to the STS action being inadvertently directed to the KMS endpoint defined in the user's configuration. This update ensures that KMS requests are appropriately routed to the custom endpoint specified by the user, while STS and other AWS services continue to utilise their default endpoints.

#### Release Note

Bug fix for [1175](https://github.com/sigstore/sigstore/issues/1175)

#### Documentation

N/A does not require a change to documentation.


haydentherapper:
Thanks @euanhume! Have you been able to use this locally and confirm that it's working as expected?

haydentherapper:
Could you rebuild Cosign locally with this change incorporated? I don't have any suggestions on how to set up AWS to test this out though.
The only thing I'm not sure about is the behavior of the client when returning `aws.Endpoint{}` without a URL set.

haydentherapper:
Closing, feel free to reopen if there's progress.

PR #1525, opened 2023-11-22
Fix cosign generate Azure key pair bug

malancas:
#### Summary

Addresses the bug reported in https://github.com/sigstore/cosign/issues/3094. That bug reports the inability to generate a key pair using the Azure Key Vault KMS. After some digging, I found the issue had to do with where the function that returns key pair's hash function is called.

This updates the Azure KMS so the key pair hash function is requested when signing and verifying, instead of retrieving it when the client is created.

This pull request also adds more logic around testing whether the a key pair with a given name already exists before attempting to generate one.


pkg/signature/kms/azure/signer.go
```diff
@@ -70,11 +70,6 @@ func LoadSignerVerifier(defaultCtx context.Context, referenceStr string) (*Signe
 		return nil, err
 	}
 
-	a.hashFunc, _, err = a.client.getKeyVaultHashFunc(defaultCtx)
```
malancas, line 73: This was causing the bug reported in https://github.com/sigstore/cosign/issues/3094. The call to `getKeyVaultHashFunc` would fail because it would try to get the hash function of a non-existent key.

haydentherapper (approved):

PR #1595, opened 2024-01-11
Add support for Ed25519ph Signer/Verifier

### LINKED ISSUE #1594: Add support for ed25519ph signer/verifier
**Description**

At Trail of Bits, we're looking at implementing part of the [Configurable Crypto Algorithms](https://docs.google.com/document/d/1-wF1t6lmJO37BzStbyz8ZkDhNm86HyvKZSaDl7uyHxI) proposal (specifically Phase 1). See https://github.com/sigstore/cosign/issues/3271, https://github.com/sigstore/rekor/issues/1724, https://github.com/sigstore/fulcio/issues/1388 for related issues in the other repos.

We started doing some work with https://github.com/sigstore/fulcio/pull/1517, which tries to support only client keys of type ecdsa+sha256 and ed25519. Fulcio does not really care about the user keys, however Rekor does and right now it does not support ed25519 (see https://github.com/sigstore/rekor/issues/1724 for details).

As suggested there and in other rekor issues, the idea is to add support for ed25519ph which is a pre-hashed version of ed25519. Having this support in [sigstore/sigstore](https://github.com/sigstore/sigstore) would allow sigstore-go, rekor, and cosign to use ed25519 keys as an alternative to the default ecdsa + sha256.

cc @tetsuo-cpp @woodruffw

ret2libc:
#### Summary

Add support for Ed25519ph digital signature to be used by other repos like Rekor, Cosign and sigstore-go.

Full e2e tests require changes in cosign and rekor as well.
Tests have been added for the Ed25519ph Signer/Verifier similar to the others.

#### Release Note

* Added `ED25519phSigner`, `ED25519phVerifier`, and `ED25519phSignerVerifier`
* Changed `LoadSigner`, `LoadVerifier`, and `LoadSignerVerifier` to default to the `ph` version of Ed25519 whenever a ed25519 private/public key is found.
* Closed #1594 

#### Documentation
No changes in documentation required, I think.


ret2libc:
I'd appreciate any comment or concern, in particular with regards to the change of behaviour of `LoadSigner`, `LoadVerifier`, and `LoadSignerVerifier`. Would that break anything currently used by rekor/cosign/sigstore-go?

pkg/signature/signerverifier.go
```diff
@@ -38,14 +38,16 @@ type SignerVerifier interface {
...
+// Ed25519SignerVerifier is desired instead, use the LoadEd25519SignerVerifier() method directly.
 func LoadSignerVerifier(privateKey crypto.PrivateKey, hashFunc crypto.Hash) (SignerVerifier, error) {
 	switch pk := privateKey.(type) {
 	case *rsa.PrivateKey:
 		return LoadRSAPKCS1v15SignerVerifier(pk, hashFunc)
 	case *ecdsa.PrivateKey:
 		return LoadECDSASignerVerifier(pk, hashFunc)
 	case ed25519.PrivateKey:
-		return LoadED25519SignerVerifier(pk)
+		return LoadED25519phSignerVerifier(pk)
```
haydentherapper, line 50: As discussed offline, we'll want to keep the default as ed25519 for signers and verifiers to avoid breaking changes across the ecosystem.

I do think this is a good opportunity to figure out the right top level method for handling the "special cases" like RSA-PSS and now ed25519ph. Given only a public or private key type for rsa or ed25519, obviously it's not possible to know which is preferred between pkcs1/pss and no-ph/ph.

Now that I'm typing this, ed25519ph suppport could be added seamlessly by assuming that specifying the hashFunc to be sha512 + a ed25519 private key, we would use the prehash variant. This still is a breaking change, but at least not a change to the API. I'd like to look across the ecosystem to see if anyone is specifying sha512 currently, but I don't think they are, or they just use crypto.Hash(0) to signal not caring about the hash func.

ret2libc: From a look around with `org:sigstore language:golang /(?-i)LoadSignerVerifier/` (and similar for `LoadSigner` and `LoadVerifier`) I did not spot any use except sha256.

ret2libc:
A few comments on the last 2 commits:
- I have used just one `SignerVerifierType` for `LoadVerifier`, `LoadSigner`, and `LoadSignerVerifier` to make sure they all handle the same types of keys. Does it make sense to you?
- I kept `SignerVerifierOptions`, `SignerOptions`, and `VerifierOptions` separate though for more flexibility. Opinions?
- Would you prefer a new `LoadSignerVerifierWithOpts` method and keeping `LoadSignerVerifier` unchanged to avoid breaking compatibility with users of the lib and allow a more gradual switch to the more flexible API?

ret2libc:
> A few comments on the last 2 commits:
> 
> * I have used just one `SignerVerifierType` for `LoadVerifier`, `LoadSigner`, and `LoadSignerVerifier` to make sure they all handle the same types of keys. Does it make sense to you?

Actually, we probably want to allow a combination of the `SignerVerifierType`, so you could so something like:
```
signature.LoadSigner(test.signer, crypto.SHA256, signature.RSAPSSSignerVerifierType | signature.ED25519phSignerVerifierType, nil)
```

haydentherapper:
@ret2libc I'm going to be taking a look at all PRs later today and tomorrow. I have a bit of a backlog I'm working through.

ret2libc:
> @ret2libc I'm going to be taking a look at all PRs later today and tomorrow. I have a bit of a backlog I'm working through.

Thanks! Actually, I have a WIP branch which I think is a bit cleaner. It uses the Options pattern and it looks like this:

```golang
func LoadSignerVerifier(privateKey crypto.PrivateKey, hashFunc crypto.Hash, opts ...SignerVerifierOption) (SignerVerifier, error) {
	o := makeSignerVerifierOpts(opts...)

	switch pk := privateKey.(type) {
	case *rsa.PrivateKey:
		if o.rsaPSSOptions != nil {
			return LoadRSAPSSSignerVerifier(pk, hashFunc, o.rsaPSSOptions)
		}
		return LoadRSAPKCS1v15SignerVerifier(pk, hashFunc)
	case *ecdsa.PrivateKey:
		return LoadECDSASignerVerifier(pk, hashFunc)
	case ed25519.PrivateKey:
		if o.useED25519ph {
			return LoadED25519phSignerVerifier(pk)
		}
		return LoadED25519SignerVerifier(pk)
	}
	return nil, errors.New("unsupported public key type")
}
```

which avoids having to modify all callers of `LoadSignerVerifier`.
A call would look like this:
```golang
sv, err := LoadSignerVerifier(privateKey, crypto.SHA256, WithED25519ph(), WithRSAPSS(opts))
```

What do you think? You can have a quick look at it at https://github.com/trail-of-forks/sigstore/tree/ed25519ph-2 and I can use that instead if you like that approach more.

haydentherapper:
@ret2libc I'm a big fan of that API design, that looks great!

Do you think we can create a second set of functions, `LoadVerifierWithOpts`, etc? I'd like to avoid breaking changes for now. Or can we have the interface use variadic `...` arguments like you proposed? I believe this would avoid the breaking change as long as the variadic parameters come last in the API.

ret2libc:
> @ret2libc I'm a big fan of that API design, that looks great!

> I'd like to avoid breaking changes for now. Or can we have the interface use variadic `...` arguments like you proposed? I believe this would avoid the breaking change as long as the variadic parameters come last in the API.

I think this is still, in theory, a breaking change as it changes the signature of the API. Probably it won't be noticed in practice, but it depends on how others are using the Signature/Verifier API.

> Do you think we can create a second set of functions, `LoadVerifierWithOpts`, etc? 

I can, yes!

pkg/signature/ed25519ph.go
```diff
@@ -0,0 +1,210 @@
...
+
+// SignMessage signs the provided message. If the message is provided,
+// this method will compute the digest according to the hash function specified
+// when the ECDSASigner was created.
+//
+// This function recognizes the following Options listed in order of preference:
+//
+// - WithDigest()
+//
+// - WithCryptoSignerOpts()
```
bobcallaway, line 57: while it does technically support this Option, I'd remove this from the doc as all other algorithms are effectively prohibited

pkg/signature/ed25519ph.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// All other options are ignored if specified.
+func (e ED25519phSigner) SignMessage(message io.Reader, opts ...SignOption) ([]byte, error) {
+	digest, _, err := ComputeDigestForSigning(message, crypto.SHA512, ed25519phSupportedHashFuncs, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return e.priv.Sign(nil, digest, &ed25519.Options{Hash: crypto.SHA512})
```
bobcallaway, line 66:
```suggestion
	return e.priv.Sign(nil, digest, crypto.SHA512)
```

pkg/signature/ed25519ph.go
```diff
@@ -0,0 +1,210 @@
...
+// All other options are ignored if specified.
+func (e *ED25519phVerifier) VerifySignature(signature, message io.Reader, opts ...VerifyOption) error {
+	digest, _, err := ComputeDigestForVerifying(message, crypto.SHA512, ed25519phSupportedHashFuncs, opts...)
+	if err != nil {
+		return err
+	}
+
+	if signature == nil {
+		return errors.New("nil signature passed to VerifySignature")
+	}
```
bobcallaway, line 142: I'd swap the order here - no reason to burn CPU cycles if we don't have a signature
```suggestion
	if signature == nil {
		return errors.New("nil signature passed to VerifySignature")
	}
	
	digest, _, err := ComputeDigestForVerifying(message, crypto.SHA512, ed25519phSupportedHashFuncs, opts...)
	if err != nil {
		return err
	}

```

pkg/signature/ed25519ph.go
```diff
@@ -0,0 +1,210 @@
...
+//
+// If opts are specified, the hash function in opts.Hash should be the one used to compute
+// digest. If opts are not specified, the value provided when the signer was created will be used instead.
+func (e ED25519phSigner) Sign(_ io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+	ed25519phOpts := []SignOption{options.WithDigest(digest)}
+	if opts != nil {
+		ed25519phOpts = append(ed25519phOpts, options.WithCryptoSignerOpts(opts))
+	}
+
+	return e.SignMessage(nil, ed25519phOpts...)
```
bobcallaway, line 96:
```suggestion
func (e ED25519phSigner) Sign(_ io.Reader, digest []byte, _ crypto.SignerOpts) ([]byte, error) {
	return e.SignMessage(nil, options.WithDigest(digest))
```

pkg/signature/signer_test.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2022 The Sigstore Authors.
```
bobcallaway, line 1:
```suggestion
// Copyright 2024 The Sigstore Authors.
```

pkg/signature/signerverifier_test.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2022 The Sigstore Authors.
```
bobcallaway, line 1:
```suggestion
// Copyright 2024 The Sigstore Authors.
```

bobcallaway (commented):
a couple nits but overall LGTM

pkg/signature/ed25519ph.go
bobcallaway: better to signal that we don't take any `crypto.SignerOpts` in the func declaration than squelch it implicitly IMO

haydentherapper: Yep, I would mirror the ed25519 signer interface - https://github.com/sigstore/sigstore/blob/main/pkg/signature/ed25519.go#L86

haydentherapper (commented):
Looks great!

ret2libc:
I think I've addressed all comments, thanks for the reviews!
Let me know if other changes are required :)

pkg/signature/ed25519ph.go
```diff
@@ -0,0 +1,201 @@
...
+// PublicKey returns the public key that is used to verify signatures by
+// this verifier. As this value is held in memory, all options provided in arguments
+// to this method are ignored.
+func (e *ED25519phVerifier) PublicKey(_ ...PublicKeyOption) (crypto.PublicKey, error) {
+	return e.publicKey, nil
+}
+
+// VerifySignature verifies the signature for the given message. Unless provided
+// in an option, the digest of the message will be computed using the hash function specified
+// when the ECDSAVerifier was created.
```
haydentherapper, line 116:
```suggestion
// when the ED25519phVerifier was created.
```

haydentherapper (dismissed):

pkg/signature/signerverifier.go
```diff
@@ -50,6 +50,28 @@ func LoadSignerVerifier(privateKey crypto.PrivateKey, hashFunc crypto.Hash) (Sig
 	return nil, errors.New("unsupported public key type")
 }
 
+// LoadSignerVerifierWithOpts returns a signature.SignerVerifier based on the
+// algorithm of the private key provided and the user's choice.
+func LoadSignerVerifierWithOpts(privateKey crypto.PrivateKey, hashFunc crypto.Hash, opts ...SignerVerifierOption) (SignerVerifier, error) {
```
ret2libc, line 55:
```suggestion
func LoadSignerVerifierWithOpts(privateKey crypto.PrivateKey, opts ...SignerVerifierOption) (SignerVerifier, error) {
```

As I had to add `opts ...SignerVerifierOption` to a bunch of methods in cosign, I wonder whether it makes sense to consider the hashFunc just part of the options, so that it can be passed more easily around.

Right now, crypto.SHA256 is hard-coded in many places and making that configurable would require adding a `hashFunc` argument to other functions (plus the SignerVerifierOption), so maybe we could just include the hash there.

What do you think?

ret2libc: I added a commit with this change. You can also see in https://github.com/trail-of-forks/cosign/compare/support-ed25519ph...trail-of-forks:cosign:signing-algorithm-flag?expand=1 how it might be used in cosign.

haydentherapper: The downside of that would be having to handle when a hash isn't provided. `isSupportedAlg` will catch if no hash is passed, but it's an incorrect error message ("incorrect hash passed" rather than "no hash passed").

haydentherapper: Ah I missed https://github.com/sigstore/sigstore/pull/1595/files#diff-caa770c16c3f35a06be96442a7e99690a11c9ff12efbffa523e054dbbed29a10R98 - This covers my concern. I thought we didn't have a fallback.

pkg/signature/options.go
```diff
@@ -55,3 +56,50 @@ type VerifyOption interface {
 	RPCOption
 	MessageOption
 }
+
+type signerVerifierOpts struct {
+	hashFunc      crypto.Hash
+	useED25519ph  bool
+	rsaPSSOptions *rsa.PSSOptions
```
haydentherapper, line 63: `PSSOptions` does include a hash function as an option. It is well-defined if both are set - https://pkg.go.dev/crypto/rsa#PSSOptions notes that what's in `PSSOptions` will override the hash function passed to it. Not a significant issue, but just something to note.

haydentherapper (approved):

pkg/signature/options.go
ret2libc: Shall we check that the Hash in PSSOptions is the same as the hash specified in `signerVerifierOpts`? We could do that check inside `makeSignerVerifierOpts`.

pkg/signature/options.go
```diff
@@ -55,3 +56,50 @@ type VerifyOption interface {
 	RPCOption
 	MessageOption
 }
+
+type signerVerifierOpts struct {
```
bobcallaway, line 60: A quick comment but will add more later - can we perhaps put these under appropriate `SignOption` and `VerifierOption` above, and then have a `SignerVerifierOption` interface that combines both?

bobcallaway: also, I'd rather keep this file as a list of interfaces, and move any specific implementations into the `options` subdirectory

ret2libc: > A quick comment but will add more later - can we perhaps put these under appropriate SignOption and VerifierOption above, and then have a SignerVerifierOption interface that combines both?

How would that work?

I'd imagine we have these functions in the interface:
```golang
	ApplyHash(*crypto.Hash)
	ApplyRSAPSSOptions(*rsa.PSSOptions)
	ApplyIsED25519ph(*bool)
```

however, where should I put them? They are not strictly used while signing and they are not strictly used while verifying, but they are used when the SignerVerifier objects are created.

> also, I'd rather keep this file as a list of interfaces, and move any specific implementations into the options subdirectory

Sure!

pkg/signature/options.go
haydentherapper: I'm fine with PSSOptions overriding hashFunc, maybe add a comment about this?

pkg/signature/options.go
bobcallaway: OK, I took a bit more time and reviewed the bulk of the code again. I think what was confusing to me initially (which you stated, but I overlooked) that these new options are relevant at initialization time only, whereas all the other options are used at the time of actual signing/verification operations.

Given that realization, I feel a bit more strongly about defining these in:
* a separate `.go` file (i.e. not in options.go)
* having a different interface name (`LoadOptions`)?

haydentherapper: I also noticed there is some overlap with `CryptoSignerOpts` too, that you can specify a hash there. I agree that splitting this out helps to distinguish between at-init and at-use options. We might want to note that there are multiple ways of specifying a hash too.

pkg/signature/signerverifier_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestLoadRSAPSSSignerVerifier(t *testing.T) {
+	opts := &rsa.PSSOptions{
+		Hash: crypto.SHA256,
+	}
+
+	privateKey, err := cryptoutils.UnmarshalPEMToPrivateKey([]byte(rsaKey), cryptoutils.SkipPassword)
+	if err != nil {
+		t.Errorf("unexpected error unmarshalling private key: %v", err)
+	}
+	sv, err := LoadSignerVerifierWithOpts(privateKey, WithHash(crypto.SHA256), WithED25519ph(), WithRSAPSS(opts))
```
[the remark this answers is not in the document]
ret2libc: I think it should be just ignored and I just wanted to test you could pass multiple configurations. I can remove it if preferred.

pkg/signature/options.go
ret2libc: Ok perfect, I will move them in separate go file! Is `options` dir still a good place for it?

@haydentherapper are you suggesting to implement `CryptoSignerOpts` as well or were you just noticing that there is an overlap?

haydentherapper: Just noticing overlap, that a hash alg can be overridden per signing event.

pkg/signature/signerverifier_options.go
```diff
@@ -0,0 +1,70 @@
+//
```
bobcallaway, line 1: can you move this file into `options` subdir please?

pkg/signature/signerverifier_test.go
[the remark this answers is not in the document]
bobcallaway: ok, makes sense

pkg/signature/ed25519ph.go
```diff
@@ -0,0 +1,201 @@
...
+	"io"
+
+	"github.com/sigstore/sigstore/pkg/signature/options"
+)
+
+var ed25519phSupportedHashFuncs = []crypto.Hash{
+	crypto.SHA512,
+}
+
+// ED25519phSigner is a signature.Signer that uses the Ed25519 public-key signature system
```
bobcallaway, line 33: worth adding this comment to denote that `ph` stands for prehashed inputs?

bobcallaway (commented):
two nits, otherwise LGTM

pkg/signature/signerverifier_options.go
ret2libc: Moving this into `options` mean changing the package to `options`. Then the `makeSignerVerifierOpts` needs to be made public, as well as the `loadOpts` structure, which would make things more rigid and harder to modify.

Is this what we want? Am I missing a nice trick to avoid this?
`LoadSignerWithOpts` (and the other functions) are in `signature` package and they should be able to easily construct the options object.

bobcallaway: why can't you define an interfaces in `signature/options.go` and then put the implementation in `signature/options/loadoptions.go`?

ret2libc: Ok I've changed the way I implement the option pattern and now it should be fine!

bobcallaway:
please fix the linting issues, otherwise LGTM

ret2libc:
Great! Thanks a lot for the review! Should be good now.

bobcallaway (approved):

PR #1601, opened 2024-01-18
Create Algorithm Registry API

tetsuo-cpp:
#### Summary

As part of our work to support configurable crypto algorithms across the Sigstore stack, we've added a `KnownSignatureAlgorithm` enumeration to `protobuf-specs` to reflect what signature algorithms are part of the official Sigstore algorithm registry. This PR adds some helpers and APIs to make use of this registry in Sigstore services.

#### Release Note

* Added `GetAlgorithmDetails` helper to access the key algorithm, hash algorithm and other algorithm-specific parameters for a given signature algorithm.
* Added `AlgorithmRegistryConfig` API to check a public key and hash algorithm combination against a set of accepted signature algorithms.
* Added `FormatSignatureAlgorithmFlag` and `ParseSignatureAlgorithmFlag` to convert the signature algorithm type to and from a string representation that can be used as a CLI parameter.


tetsuo-cpp:
I'm working on some unit tests now.

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,212 @@
...
+	// GetSignatureAlgorithm returns the algorithm registry.
+	GetSignatureAlgorithm() v1.KnownSignatureAlgorithm
+
+	// GetKeyType returns the public key algorithm for the given signature algorithm.
+	GetKeyType() PublicKeyType
+
+	// GetHashType returns the hash algorithm for a given signature algorithm.
+	GetHashType() crypto.Hash
+
+	// GetRSAKeySize returns the key size if the public key algorithm is RSA.
```
tetsuo-cpp, line 52: @ret2libc This is how I'm exposing parameters that are specific to each public key.

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,212 @@
...
+	RSA     PublicKeyType = iota // RSA public key
+	ECDSA                        // ECDSA public key
+	ED25519                      // ED25519 public key
+)
+
+// RSAKeySize represents the size of an RSA public key.
+type RSAKeySize int
+
+// AlgorithmDetails exposes relevant information for a given signature algorithm.
+type AlgorithmDetails interface {
```
tetsuo-cpp, line 42: I made this interface because I don't think it makes sense for users to be constructing their own.

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,212 @@
...
+	return rsaKeySize, nil
+}
+
+func (a algorithmDetailsImpl) GetECDSACurve() (*elliptic.Curve, error) {
+	if a.keyType != ECDSA {
+		return nil, fmt.Errorf("unable to retrieve ECDSA curve for key type: %T", a.keyType)
+	}
+	ecdsaCurve, ok := a.extraKeyParams.(elliptic.Curve)
+	if !ok {
+		panic("unable to retrieve curve for ECDSA, malformed algorithm details?")
```
tetsuo-cpp, line 101: I'm using `panic` pretty freely here since it's all dependent on what's in `algorithmDetails` and therefore, there shouldn't be any way for the user to reach this.

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,212 @@
...
+		if err != nil {
+			return nil, err
+		}
+		permittedAlgorithms = append(permittedAlgorithms, a)
+	}
+	return &AlgorithmRegistryConfig{permittedAlgorithms: permittedAlgorithms}, nil
+}
+
+// IsAlgorithmPermitted checks whether a given public key/hash algorithm combination is permitted by a registry config.
+func (registryConfig AlgorithmRegistryConfig) IsAlgorithmPermitted(key crypto.PublicKey, hash crypto.Hash) error {
```
tetsuo-cpp, line 183: Perhaps this should just return `bool`.

tetsuo-cpp:
CC: @ret2libc @woodruffw @haydentherapper

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,212 @@
...
+//
+// Individual services may wish to restrict what algorithms are allowed to a subset of what is covered in the algorithm
+// registry (represented by v1.KnownSignatureAlgorithm).
+type AlgorithmRegistryConfig struct {
+	permittedAlgorithms []AlgorithmDetails
+}
+
+// GetAlgorithmDetails retrieves a set of details for a given v1.KnownSignatureAlgorithm flag that allows users to
+// introspect the public key algorithm, hash algorithm and more.
+func GetAlgorithmDetails(knownSignatureAlgorithm v1.KnownSignatureAlgorithm) (AlgorithmDetails, error) {
```
ret2libc, line 160: Does this need to be public?

pkg/signature/algorithm_registry.go
ret2libc: Do you think users will ever need to access this? I don't see functions that return an `AlgorithmDetails` (apart from `GetAlgorithmDetails` which I think can be made private).

I'd be more for keeping this private unless we see a real use case for having all these info. It should also be easier to add new public APIs in the future if we need them rather than modifying already public interfaces/structures which would require major versions updates.

ret2libc: Of course if we do make it private we could just remove it altogether and keep only the `Impl` struct for now.

woodruffw: Yes, I *think* we can make this private for now.

pkg/signature/algorithm_registry.go
woodruffw: I think not, assuming you don't need it 🙂

pkg/signature/algorithm_registry.go
woodruffw: `bool` makes sense to me -- the only error here is "not permitted," so there's nothing to signal that a `bool` can't signal.

pkg/signature/algorithm_registry.go
tetsuo-cpp: I thought we agreed that we wanted to expose a way of getting the hash function, key type, etc for a given signature algorithm, unless I misunderstood that conversation.

ret2libc: Thinking this a bit more and you are probably right @tetsuo-cpp , we might need this after all. At least, it sounds like a good type to move around between functions that need to know what kind of signing algorithm needs to be applied.

For example, in cosign there is this code:
```golang
func signerFromNewKey() (*SignerVerifier, error) {
	privKey, err := cosign.GeneratePrivateKey()
	if err != nil {
		return nil, fmt.Errorf("generating cert: %w", err)
	}
	sv, err := signature.LoadECDSASignerVerifier(privKey, crypto.SHA256)
	if err != nil {
		return nil, err
	}

	return &SignerVerifier{
		SignerVerifier: sv,
	}, nil
}
```

And it will likely need a new argument (coming down from the CLI flag) that specifies which type of private key to use (and its arguments) and the hash function to use with the Signer/Verifier.

So that could probably become:
```golang
func signerFromNewKey(algorithmDetails AlgorithmDetails) (*SignerVerifier, error) {
	privKey, err := cosign.GeneratePrivateKey(algorithmDetails)
	if err != nil {
		return nil, fmt.Errorf("generating cert: %w", err)
	}
	sv, err := signature.LoadSignerVerifier(privKey, algorithmDetails.GetHashType())
	if err != nil {
		return nil, err
	}

	return &SignerVerifier{
		SignerVerifier: sv,
	}, nil
}
```
and then the `GeneratePrivateKey` can do something like this:
```golang
	switch signingAlgorithm.GetKeyType() {
	case signature.ECDSA:
		curve, err := signingAlgorithm.GetECDSACurve()
		if err != nil {
			return nil, err
		}
		return ecdsa.GenerateKey(*curve, rand.Reader)
	case signature.ED25519:
		_, priv, err := ed25519.GenerateKey(rand.Reader)
		return priv, err
	default:
		return nil, fmt.Errorf("unsupported signing algorithm: %s", signingAlgorithm)
	}
```

ret2libc:
@tetsuo-cpp you can see https://github.com/trail-of-forks/cosign/compare/support-ed25519ph...trail-of-forks:cosign:signing-algorithm-flag?expand=1 on how I might be using the registry in cosign. Hope it helps us define a good API!

woodruffw:
(Force-pushed to fix the signoffs.)

pkg/signature/algorithm_registry.go
haydentherapper: I prefer to avoid panics, just in case some refactor occurs and this ends up being user-controlled (contrived, but possible). Given the API returns an error already, I'd just return an error.

I see there's a few places below that don't return an error, we might just need to plumb errors through.

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// GetECDSACurve returns the curve if the public key algorithm is ECDSA.
+	// Otherwise, an error is returned.
+	GetECDSACurve() (*elliptic.Curve, error)
+
+	checkKey(crypto.PublicKey) bool
+	checkHash(crypto.Hash) bool
+}
+
+type algorithmDetailsImpl struct {
```
haydentherapper, line 64: Can you add some comments about what each of these fields represent, particularly `flagValue` and `extraKeyParams`?

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,239 @@
...
+		if err != nil {
+			return nil, err
+		}
+		permittedAlgorithms = append(permittedAlgorithms, a)
+	}
+	return &AlgorithmRegistryConfig{permittedAlgorithms: permittedAlgorithms}, nil
+}
+
+// IsAlgorithmPermitted checks whether a given public key/hash algorithm combination is permitted by a registry config.
+func (registryConfig AlgorithmRegistryConfig) IsAlgorithmPermitted(key crypto.PublicKey, hash crypto.Hash) bool {
```
haydentherapper, line 183: How do you foresee this being used with KMS keys? Should the caller be required to inspect the KMS key for its key and hash (we have key, we're missing hash, but it's being implemented in https://github.com/sigstore/sigstore/pull/1426). Should each KMS provider has its own registry?(as a service, this would be a bit annoying to use, since you'd have to consult a registry of regstries)

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,239 @@
...
+
+// HasAlgorithmDetails checks whether a given v1.KnownSignatureAlgorithm flag is permitted by the registry.
+func (registryConfig AlgorithmRegistryConfig) HasAlgorithmDetails(signatureAlgorithm v1.KnownSignatureAlgorithm) bool {
+	_, err := registryConfig.GetAlgorithmDetails(signatureAlgorithm)
+	return err == nil
+}
+
+// FormatSignatureAlgorithmFlag formats a v1.KnownSignatureAlgorithm to a string that conforms to the conventions of CLI
+// arguments that are used for Sigstore services.
+func FormatSignatureAlgorithmFlag(algorithm v1.KnownSignatureAlgorithm) (*string, error) {
```
haydentherapper, line 221: Any reason for a string pointer over a string?

tetsuo-cpp: Just so I don't have to construct a string in the error case. I can just create an empty one though if that's preferable.

pkg/signature/algorithm_registry.go
tetsuo-cpp: Sure, that should be easy to do.

pkg/signature/algorithm_registry.go
tetsuo-cpp: I think it makes most sense for the calling code to inspect the KMS key and hash, and check against the registry to see whether it should use that key or not.

I don't think it makes sense for each KMS provider to own its own registry, but perhaps the `kms.SignerVerifier` API could take an optional `AlgorithmRegistryConfig` and do the checking there? But I'd lean towards not tangling the APIs together like that and just let the calling code compose these calls as required.

What are your thoughts?

pkg/signature/algorithm_registry.go
tetsuo-cpp: [12edb0f](https://github.com/sigstore/sigstore/pull/1601/commits/12edb0f94732c78348507edbb297d9330536511b)

pkg/signature/algorithm_registry.go
tetsuo-cpp: [1d74008](https://github.com/sigstore/sigstore/pull/1601/commits/1d74008e9c24c141132f550f38e158455e46c259)

pkg/signature/algorithm_registry.go
tetsuo-cpp: [021f385](https://github.com/sigstore/sigstore/pull/1601/commits/021f385ab7e86b9c2f180e11f266f6855e0b122d)

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,259 @@
...
+		// This should be unreachable.
+		return nil, fmt.Errorf("unable to retrieve curve for ECDSA, malformed algorithm details?: %T", a.keyType)
+	}
+	return &ecdsaCurve, nil
+}
+
+func (a algorithmDetailsImpl) checkKey(pubKey crypto.PublicKey) (bool, error) {
+	switch a.keyType {
+	case RSA:
+		rsaKey, ok := pubKey.(rsa.PublicKey)
```
ret2libc, line 125:
```suggestion
		rsaKey, ok := pubKey.(*rsa.PublicKey)
```

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,259 @@
...
+		if !ok {
+			return false, nil
+		}
+		keySize, err := a.GetRSAKeySize()
+		if err != nil {
+			return false, err
+		}
+		return rsaKey.Size() == int(keySize), nil
+	case ECDSA:
+		ecdsaKey, ok := pubKey.(ecdsa.PublicKey)
```
ret2libc, line 135:
```suggestion
		ecdsaKey, ok := pubKey.(*ecdsa.PublicKey)
```

ret2libc (changes requested):

pkg/signature/algorithm_registry.go
tetsuo-cpp: [a3d6a56](https://github.com/sigstore/sigstore/pull/1601/commits/a3d6a56055275e23623771275148eff10dfe4b19)

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0
 	github.com/segmentio/ksuid v1.0.4
+	github.com/sigstore/protobuf-specs v0.2.2-0.20240118022641-cea4265fd6c5
```
haydentherapper, line 16: Can this be v0.3.0rc0? Or is that not a valid version in golang?

pkg/signature/algorithm_registry.go
haydentherapper: That sounds good to me. Long-term I'd lean towards the proposal to take in a config, just to avoid each client having to reimplement this, but I don't think it's something that's blocking regardless, just thinking through how we'll handle users who provide their own KMS keys.

haydentherapper (dismissed):

go.mod
tetsuo-cpp: When I add that tag to `go.mod`, `go mod tidy` corrects it to this version. Judging from [this](https://go.dev/doc/modules/version-numbers), it probably should be v0.3.0-rc0 rather than v0.3.0rc0.

woodruffw: Hmm, this is something I can probably unilaterally fix -- I can push up a `v0.3.0-rc1` tag to protobuf-specs. Once that's in, you should be able to bump here.

woodruffw: @tetsuo-cpp I've created `v0.3.0-beta.2`: https://github.com/sigstore/protobuf-specs/releases/tag/v0.3.0-beta.2

tetsuo-cpp: Awesome, I'll use that tag.

tetsuo-cpp: [c14bfce](https://github.com/sigstore/sigstore/pull/1601/commits/c14bfce5f9b2b5018574faae4b8fa5cb7474608f)

haydentherapper (dismissed):

ret2libc (approved):

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,259 @@
...
+// Individual services may wish to restrict what algorithms are allowed to a subset of what is covered in the algorithm
+// registry (represented by v1.KnownSignatureAlgorithm).
+type AlgorithmRegistryConfig struct {
+	permittedAlgorithms []AlgorithmDetails
+}
+
+// GetAlgorithmDetails retrieves a set of details for a given v1.KnownSignatureAlgorithm flag that allows users to
+// introspect the public key algorithm, hash algorithm and more.
+func GetAlgorithmDetails(knownSignatureAlgorithm v1.KnownSignatureAlgorithm) (AlgorithmDetails, error) {
+	for _, detail := range algorithmDetails {
```
bobcallaway, line 177: maybe turn this into a map[v1.KnownSignatureAlgorithm]AlgorithmDetails ?

tetsuo-cpp: I was thinking about this, but my reasons not to were:
- If I do that, I have to match the `v1.KnownSignatureAlgorithm` in the `AlgorithmDetails` with the one in the map key, which is another opportunity for subtle typos that can introduce bugs. I can't remove it from `AlgorithmDetails` since we need it for things like `AlgorithmRegistryConfig.ListPermittedAlgorithms`.
- The `algorithmDetails` array only includes an entry for each signature algorithm in the registry which by design, is a small, constrained set of algorithms. I haven't benchmarked and to be honest, I don't know much about Go's performance characteristics, but I wouldn't expect the map to be faster than a simple linear search for this number of elements.
- This array isn't part of the API so if my assumptions are wrong, we can easily walk this decision back if we need to without introducing breaking changes.

Happy to change it over to a map, if you'd prefer though.

pkg/signature/algorithm_registry_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Errorf("unexpected success getting rsa key size")
+	}
+}
+
+func TestAlgorithmRegistryConfig(t *testing.T) {
+	config, err := NewAlgorithmRegistryConfig([]v1.KnownSignatureAlgorithm{v1.KnownSignatureAlgorithm_ECDSA_SHA2_256_NISTP256, v1.KnownSignatureAlgorithm_ED25519})
+	if err != nil {
+		t.Errorf("unexpected error creating algorithm registry config: %v", err)
+	}
+
```
haydentherapper, line 62: Can we also add a test for RSA keys being permitted?

It might be overkill, but ideally we'd have a test for every supported key type, to have maximum code coverage.

haydentherapper (commented):
Overall LGTM, please take a look at the lint failure

pkg/signature/algorithm_registry_test.go
tetsuo-cpp: [c7acffd](https://github.com/sigstore/sigstore/pull/1601/commits/c7acffd09c81246401a30270fe6605d18c5713ea)

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,259 @@
...
+// PublicKeyType represents the public key algorithm for a given signature algorithm.
+type PublicKeyType uint
+
+const (
+	RSA     PublicKeyType = iota // RSA public key
+	ECDSA                        // ECDSA public key
+	ED25519                      // ED25519 public key
+)
+
+// RSAKeySize represents the size of an RSA public key.
```
haydentherapper, line 38:
```suggestion
// RSAKeySize represents the size of an RSA public key in bits.
```

haydentherapper:
@kommendorkapten Do you want to take a look at this PR? There's an open question of how this should interact with `PublicKeyDetails` ([thread](https://github.com/sigstore/protobuf-specs/pull/199#issuecomment-1909460347))

kommendorkapten:
This is great 🚀  
This is quite related to the existing [PublicKeyDetails](https://github.com/sigstore/protobuf-specs/blob/cac2427a7d2dcd564806ad2e67a7dd801e31c3e0/protos/sigstore_common.proto#L52), which today is used when distributing the trust root, to indicate how to interpret and use public key material, e.g. the [public key for Rekor](https://github.com/sigstore/root-signing/blob/87a6c3ff0d82feeb1869c11574e83bbdd05e5dbc/targets/trusted_root.json#L9). 

I believe there is an overlap between `KnownSignatureAlgorithm` and `PublicKeyDetails`, and I fear this may become a bit complicated for clients to work with. How do we feel about merging them into `PublicKeyDetails`? 

`PublicKeyDetails` captures the signature algorithm used (albeit a bit underspecified for RSA keys, but that can easily be fixed), which is exactly what `KnownSignatureAlgorithm` does. 

The only differences I see is the naming, and that `PublicKeyDetails` also indicates the encoding of the public key material, but I would think that it wouldn't really matter, I would say that it's only RSA keys that in practice may have different encodings (PKCS vs PKIX).

One can argue that the two types might solve slightly different problems, but as the overlap is so big, I would feel better if they could be combined, otherwise the clients would need to translate between the two formats, which I feel is sub-optimal.

Thoughts?

tetsuo-cpp:
> The only differences I see is the naming, and that PublicKeyDetails also indicates the encoding of the public key material, but I would think that it wouldn't really matter, I would say that it's only RSA keys that in practice may have different encodings (PKCS vs PKIX).

Aren't the RSA values also missing the hash algorithm, or is that implied somehow? That was the main reason I didn't add to `PublicKeyDetails` (as well as the key encoding). It seemed to me that the `KnownSignatureAlgorithm` enum would be capturing a slightly different set of information than what's in `PublicKeyDetails`.

If it's fine to overwrite the values in `PublicKeyDetails` with what's in `KnownSignatureAlgorithm` (that is, if we can avoid distinguishing how the key is encoded as well as add hash functions and key sizes for each of the RSA options), then I don't see anything stopping us from merging the two.

kommendorkapten:
> Aren't the RSA values also missing the hash algorithm, or is that implied somehow?

I don't remember the reason we left the RSA values underspecified 🤔 , but adding new values with them in there would make no harm.

Yeah, adding the new values missing in `PublicKeyDetails` is preferable IMHO. The old RSA values can be marked as `[deprecated = true]`. I believe RSA are the only ones that needs updates, the EC types matches what's in `KownSignatureAlgorithms` (but only P256 is listed). I would also be tempted to mark deterministic ECDSA as deprecated, I don't think anyone is using that?

tetsuo-cpp:
> Yeah, adding the new values missing in PublicKeyDetails is preferable IMHO. The old RSA values can be marked as [deprecated = true]. I believe RSA are the only ones that needs updates, the EC types matches what's in KownSignatureAlgorithms (but only P256 is listed). I would also be tempted to mark deterministic ECDSA as deprecated, I don't think anyone is using that?

That seems reasonable.

I had another idea though: what do you think about marking the entire `PublicKeyDetails` enum as deprecated and then move any code using this enum to use `KnownSignatureAlgorithms`? That way, we can maintain backwards compatibility as you're suggesting without diverging from the [algorithm registry markdown](http://github.com/sigstore/protobuf-specs/blob/main/docs/algorithm-registry.md) and having lots of deprecated options mixed in with legitimate ones in the short term.

kommendorkapten:
> I had another idea though: what do you think about marking the entire PublicKeyDetails enum as deprecated 

My gut feeling is that this will be quite disruptive for the community, as this field is being used in quite a few places. And it would also require us to have mechanism of indicating the encoding of the key (especially for RSA keys). There are quite a few implementations out there already (Go, Java, JavaScript, Python, and some tooling for managing the trust root which depends on the pubplic key type).

> without diverging from the [algorithm registry markdown](https://github.com/sigstore/protobuf-specs/blob/main/docs/algorithm-registry.md)

Hm, why would the public key type diverge from that? Because of the added encoding info?

tetsuo-cpp:
> My gut feeling is that this will be quite disruptive for the community, as this field is being used in quite a few places. And it would also require us to have mechanism of indicating the encoding of the key (especially for RSA keys). There are quite a few implementations out there already (Go, Java, JavaScript, Python, and some tooling for managing the trust root which depends on the pubplic key type).

I think I understand. My reasoning was that if we're deprecating most of the existing values, it'll already be disruptive and will require modification to any code that reads/writes those enums. But there's probably some code where the enum gets passed around without operating on it which would recompile fine if we stick with `PublicKeyDetails`, whereas removing that enum will break that code. Is that what you mean?

> Hm, why would the public key type diverge from that? Because of the added encoding info?

Yeah, I expect the number of values for RSA will double since we'll be duplicating them to include the encoding info.

woodruffw:
(Sorry, catching up here!)

@kommendorkapten's proposal looks reasonable to me -- it'd be nice to deduplicate here and reuse `PublicKeyDetails` as the registry enumeration, provided that doesn't cause issues for clients (I don't see why it would).

Two thoughts:

1. Specifying the full suite (i.e. PK type + hash) shouldn't be an issue from an encoding perspective -- there are standard X.509 AlgorithmIdentifiers for RSA+PSS and RSA+PKCS1v1.5 that encode hash algorithm choices. That being said, those identifiers *don't* encode public modulus length, so in practice there won't be any X.509 encoding difference between `PKIX_RSA_PKCS1_2048_SHA256` and `PKIX_RSA_PKCS1_3072_SHA256`, etc. This may or may not matter; we may want to keep them as separate enum variants regardless 🙂 

3. The `PublicKeyDetails` name is slightly misleading, since the enum now (fully) captures both the public key type and its signature parameter set. This is *fine* IMO (especially since some variants already do this, but I figure it's worth calling it).

kommendorkapten:
See https://github.com/sigstore/protobuf-specs/pull/212 for PR on merge of `PublicKeyDetails` and `KnownSignatureAlgorithms`.

tetsuo-cpp:
> See [sigstore/protobuf-specs#212](https://github.com/sigstore/protobuf-specs/pull/212) for PR on merge of `PublicKeyDetails` and `KnownSignatureAlgorithms`.

Nice work, thanks for that. Once a release is cut, I'll get this PR up to date.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0
 	github.com/segmentio/ksuid v1.0.4
+	github.com/sigstore/protobuf-specs v0.3.0-beta.2
```
haydentherapper, line 16: Can bump up now

haydentherapper (dismissed):
Overall LGTM. As a meta comment, my only reservation is it's hard to know if we've gotten the API right here without seeing this in use. I'm not really concerned though since I don't expect this API to be used outside of the agility work initially.

ret2libc:
> Overall LGTM. As a meta comment, my only reservation is it's hard to know if we've gotten the API right here without seeing this in use. I'm not really concerned though since I don't expect this API to be used outside of the agility work initially.

@haydentherapper you can see its use in other pending PRs like https://github.com/sigstore/rekor/pull/1974/ , https://github.com/sigstore/cosign/pull/3497 , https://github.com/sigstore/fulcio/pull/1517 (this last one might be outdated)

go.mod
ret2libc: Done. I also had to use `PublicKeyDetails` as `KnownSignatureAlgorithm` was removed with recent changes.

pkg/signature/algorithm_registry.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (a algorithmDetailsImpl) checkHash(hashType crypto.Hash) bool {
+	return a.hashType == hashType
+}
+
+var algorithmDetails = []algorithmDetailsImpl{
+	{v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_2048_SHA256, RSA, crypto.SHA256, RSAKeySize(2048), "rsa-sign-pkcs1-2048-sha256"},
+	{v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_3072_SHA256, RSA, crypto.SHA256, RSAKeySize(3072), "rsa-sign-pkcs1-3072-sha256"},
+	{v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_4096_SHA256, RSA, crypto.SHA256, RSAKeySize(4096), "rsa-sign-pkcs1-4096-sha256"},
```
kommendorkapten, line 158: I think we should add the RSA/PSS options here too.

kommendorkapten: See https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L74

haydentherapper: We should also document we aren't handling the deprecated options, namely a PKCS1v1.5 encoded RSA key, preferring PKIX encoding.

ret2libc: Done!

ret2libc: I can't hide/resolve this comment, but this should be done in current version!

ret2libc (commented):
I have squashed the whole history into 1 commit because there were very old iterations that did not make sense anymore.

The PR was tested with fulcio/rekor/cosign.

With regards to the API and its uses: to be honest a few functions are not used anywhere, so we may make the API even smaller and hide more things, if we want. So if/when it grows/changes it will be easier to do so.

In particular, these functions are not used in any of the other components:
- HasAlgorithmDetails
- ListPermittedAlgorithms
- registry.GetAlgorithmDetails
- GetAlgorithmDetails

ret2libc:
ping @haydentherapper 

There has been a lot of back and forth on this PR and the others. I have updated all of them, so I'd appreciate a new review on this.

haydentherapper:
Thanks @ret2libc for the updates! I did a review over the PR and reviewed my old comments to remember where we had ended - I think the PR is looking good! 

For those functions, if they're unused, I'd agree that we can omit then. I'd rather keep the API surface smaller and expand it later as needed. Do you recall why we had proposed those? It seems like those are mostly focused on exposing the configuration mapping, which I assume would only be relevant to developers who could inspect the source code.

ret2libc:
> Thanks @ret2libc for the updates! I did a review over the PR and reviewed my old comments to remember where we had ended - I think the PR is looking good!
> 
> For those functions, if they're unused, I'd agree that we can omit then. I'd rather keep the API surface smaller and expand it later as needed. Do you recall why we had proposed those? It seems like those are mostly focused on exposing the configuration mapping, which I assume would only be relevant to developers who could inspect the source code.

I think they were just exposed to have a "nice API" and maybe also because we did not know how we would have ended up using the API when we started with this.

I'm going to clean the API!

haydentherapper (dismissed):
This looks great! Any other changes or is this good to merge?

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/secure-systems-lab/go-securesystemslib v0.9.0
 	github.com/segmentio/ksuid v1.0.4
+	github.com/sigstore/protobuf-specs v0.3.3
```
bobcallaway, line 16: can you please bump this to v0.4.0 ahead of merging?

bobcallaway (approved):

haydentherapper (approved):

haydentherapper:
Going to merge now, thanks all!

PR #1616, opened 2024-01-30
Convert ED25519phSignerVerifier to the Pure version

ret2libc:
#### Summary

In https://github.com/sigstore/fulcio/issues/1388#issuecomment-1915809491 we agreed to just use ED25519 (Pure) in Fulcio. That means that in Cosign (maybe elsewhere?) we need to convert a ED25519phSignerVerifier to the Pure version, but for doing so we need access to the Private key, thus the new `ConvertToED25519` API.

#### Release Note

* Added `ConvertToED25519` method to `ED25519phSignerVerifier` to convert the `ED2551ph` SignerVerifier to a `ED25519` one.

### DIFF
--- a/pkg/signature/ed25519ph.go
+++ b/pkg/signature/ed25519ph.go
@@ -46,6 +46,16 @@ func LoadED25519phSigner(priv ed25519.PrivateKey) (*ED25519phSigner, error) {
 	}, nil
 }
 
+// ToED25519SignerVerifier creates a ED25519SignerVerifier from a ED25519phSignerVerifier
+//
+// Clients that use ED25519phSignerVerifier should use this method to get a
+// SignerVerifier that uses the same ED25519 private key, but with the Pure
+// Ed25519 algorithm. This might be necessary to interact with Fulcio, which
+// only supports the Pure Ed25519 algorithm.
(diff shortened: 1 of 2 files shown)


ret2libc:
See https://github.com/sigstore/cosign/pull/3479/commits/16c0256645368419359401f64744357438befd17 for how this is going to be used in Cosign.

pkg/signature/ed25519ph.go
```diff
@@ -46,6 +46,11 @@ func LoadED25519phSigner(priv ed25519.PrivateKey) (*ED25519phSigner, error) {
 	}, nil
 }
 
+// ConvertToED25519 creates a ED25519SignerVerifier from a ED25519phSignerVerifier
```
haydentherapper, line 49: Can you add some more details to this comment about why we need this function?

pkg/signature/ed25519ph.go
```diff
@@ -46,6 +46,11 @@ func LoadED25519phSigner(priv ed25519.PrivateKey) (*ED25519phSigner, error) {
...
+func (e ED25519phSignerVerifier) ConvertToED25519() (*ED25519SignerVerifier, error) {
```
haydentherapper, line 50: nit: I'd call this `ToED25519SignerVerifier` to be specific with which interface it returns

haydentherapper (approved):

PR #484, opened 2022-06-02
Autoclose OAuth success page after 5 seconds.

wlynch:
#### Summary

Small QoL improvement to clean up success pages after they are no longer
needed.

Shout out to @bobcallaway for the idea!

Co-authored-by: Bob Callaway <bcallaway@google.com>
Signed-off-by: Billy Lynch <billy@chainguard.dev>

#### Ticket Link

Part of #426 

#### Release Note

```release-note
Interactive OAuth success page will now autoclose after 5 seconds.
```

PR #1618, opened 2024-02-03
Add support for autoclosing oauth flow window

stgarf:
#### Summary

This solves a small UX friction point. It can be tested by using something that makes use of the interactive flow like `gitsign` and I've integrated those changes [here](https://github.com/sigstore/gitsign/compare/main...stgarf:gitsign:feature/autoclose-auth-page) to test with.

image

Update `go.mod` to contain:
`replace github.com/sigstore/sigstore => github.com/stgarf/sigstore 510ce9e`

and then run `go mod tidy`. You should be able to test out the changes. _Of note_ is that the page will not close automatically if no connectorid (for gitsign for example) is set due to the fact that the window.close API call will only succeed if the page is the first item in the window's history.

```bash
git config --local commit.gpgsign true  # Sign all commits
git config --local tag.gpgsign true  # Sign all tags
git config --local gpg.x509.program gitsign  # Use gitsign for signing
git config --local gpg.format x509  # gitsign expects x509 args
git config --local gitsign.autoclose true
git config --local gitsign.autoclosetimeout 5
git config --local gitsign.connectorid https://accounts.google.com
```

#### Release Note

Updated the `interactiveIDTokenSource` struct to have `autoclose` and `autocloseTimeout` fields which default to `false` and `0` respectively. The original `InteractiveSuccessHTML` returned in `flow.go` has been changed from a static string to a dynamic go template which can be returned with a call to `GetInteractiveSuccessHTML(autoclose bool, timeout int)`. 

#### Documentation

This may require an update to docs but I am not sure where. I've only seen `gitsign` using this so far and I've already made some documentation updates in my fork of it.

### DIFF
--- a/pkg/oauth/interactive.go
+++ b/pkg/oauth/interactive.go
@@ -15,8 +15,120 @@
 // Package oauth implements OAuth/OIDC support for device and token flows
 package oauth
 
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+)
+
+// GetInteractiveSuccessHTML is the page displayed upon success when using a web browser during an interactive Oauth token flow.
+// The page will close automatically if autoclose is true with the timeout specified.
+func GetInteractiveSuccessHTML(autoclose bool, timeout int) (string, error) {
+	const successTemplate = `<!DOCTYPE html>
+<html>
+	<head>
+		<title>Sigstore Authentication</title>
+		<link id="favicon" rel="icon" type="image/svg"/>
+		<style>
+			:root { font-family: "Trebuchet MS", sans-serif; height: 100%; color: #444444; overflow: hidden; }
+			body { display: flex; justify-content: center; height: 100%; margin: 0 10%; background: #FFEAD7; }
+			.container { display: flex; flex-direction: column; justify-content: space-between; }
+			.sigstore { color: #2F2E71; font-weight: bold; }
+			.header { position: absolute; top: 30px; left: 22px; }
+			.title { font-size: 3.5em; margin-bottom: 30px; animation: 750ms ease-in-out 0s 1 show; }
+			.content { font-size: 1.5em; animation: 250ms hide, 750ms ease-in-out 250ms 1 show; }
+			.anchor { position: relative; }
+			.links { display: flex; justify-content: space-between; font-size: 1.2em; padding: 60px 0; position: absolute; bottom: 0; left: 0; right: 0; animation: 500ms hide, 750ms ease-in-out 500ms 1 show; }
+			.link { color: #444444; text-decoration: none; user-select: none; }
+			.link:hover { color: #6349FF; }
+			.link:hover>.arrow { transform: scaleX(1.5) translateX(3px); }
+			.link:hover>.sigstore { color: inherit; }
+			.link, .arrow { transition: 200ms; }
+			.arrow { display: inline-block; margin-left: 6px; transform: scaleX(1.5); }
+			@keyframes hide { 0%, 100% { opacity: 0; } }
+			@keyframes show { 0% { opacity: 0; transform: translateY(40px); } 100% { opacity: 1; } }
+		</style>
+	</head>
+	<body>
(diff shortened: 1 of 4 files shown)


kommendorkapten:
Code looks good, but I haven't tested it.

pkg/oauth/interactive.go
```diff
@@ -74,7 +90,40 @@ const (
...
+		</script>
+		{{- end }}
 	</body>
 </html>
 `
-)
+	// Parse the template
+	tmpl, err := template.New("success").Parse(successTemplate)
+	if err != nil {
+		log.Panicf("error parsing success template: %v", err)
```
haydentherapper, line 113: Is there another way we can structure this to avoid the panics? Return an error?

haydentherapper (changes requested):

pkg/oauth/interactive.go
stgarf: Sure thing, let me look over how errors are handled elsewhere and do similar.

stgarf: @haydentherapper I took a look at how I might accomplish this but was slightly deterred from refactoring it to return an error because I wasn't sure how it could be handled in flow.go with creating an init function to call the `GetInteractiveSuccessHTML` to get the html. I'll make a commit with that kind of change and you can let me know if you want to move forward like that or revert to the previous state.

haydentherapper:
@wlynch @imjasonh pinging you here too, @bobcallaway was wondering if you had tried this out and run into an issue

pkg/oauth/interactive.go
```diff
@@ -15,9 +15,121 @@
...
+		return "", fmt.Errorf("error executing template: %w", err)
+	}
+	return htmlPage.String(), nil
+}
+
 const (
-	// InteractiveSuccessHTML is the page displayed upon success when using a web browser during an interactive Oauth token flow.
-	InteractiveSuccessHTML = `<!DOCTYPE html>
+	// DefaultInteractiveSuccessHTML is the page displayed upon success when using a web browser during an interactive Oauth token flow.
+	DefaultInteractiveSuccessHTML = `<!DOCTYPE html>
```
wlynch, line 132: This may break downstream users that are using on the old var. If you want to rename this, perhaps we should deprecate + alias this instead?

e.g.

```go
const (
  // Deprecated: ...
  InteractiveSuccessHTML = ...
)

var (
  DefaultInteractiveSuccessHTML = InteractiveSuccessHTML
)
```

Alternatively we could only deprecate and point them to GetInteractiveSuccessHTML.

wlynch (commented):
[I tried to get it working awhile back and had trouble](https://github.com/sigstore/sigstore/pull/484#issuecomment-1150226787), but my javascript is also very weak. (@stgarf If you happened know why my last attempt to get this to work wasn't successful, I'd love to learn why! 🙏 )

@stgarf made a similar change in `chainctl` and we confirmed it works, so I'm all for this! 🚀

pkg/oauth/interactive.go
stgarf: Sure, I made that change [based on the last review](https://github.com/sigstore/sigstore/pull/1618#discussion_r1485233325) but wasn't super happy with how I ended up refactoring it. Would you like me to implement your suggestion as well?

haydentherapper: I agree on avoiding breaking changes. I am happy with the change to switch from panic to error too.

stgarf:
@wlynch to answer your question for posterity `the page will not close automatically if it was redirected to from another page, e.g. no social login was set, due to the fact that the window.close API call will only succeed if the page is the first item in the window's history.`

haydentherapper (approved):

stgarf:
Thanks folks! Commits have been squashed.

hectorj2f (approved):

haydentherapper (approved):

wlynch (approved):

PR #1638, opened 2024-02-26
Add Akeyless vaultless platform support for sigstore

rani-sharim:
Akeyless is a SaaS vaultless key management platform. 
This pr adds the ability to use Akeyless stored keys as signing keys.

The default algorithm used is `RSA PKCS1.5`, however, `ECDSA256` is also supported. 

For more info see
https://www.akeyless.io/

#### Documentation
To use:
1. Set configuration file in /var/akeyless/conf/sigstore.conf 
```
akeyless_url="http://your.akeyless.gateway.url:8081"
[auth]
access_id="<your access key>"
access_key="<your access id>"
access_type="access_key"
```

2. When setting key paths, use `akeyless://<key name>` as the key path.


haydentherapper:
Hey, thanks for the PR, and @akljph, sorry for not writing back via email.

The concern with taking on additional platforms is that the maintainers don't have familiarity with the provider, so as issues arise, debugging won't be trivial. At the same time, I would like to find a solution to unblock you and other new providers.

After chatting with maintainers, one thought we had was using a plugin system like https://github.com/hashicorp/go-plugin/blob/main/README.md. I plan to spend some time this week to see if it would be a good fit. At a glance, it seems quite promising - we create an interface for signing and verification that each provider would define, and then could load in plugins dynamically rather than bundling them in the package. If you have other suggestions too, I'd be happy to chat more.

akljph:
Thank you very much for the reply @haydentherapper. We really appreciate your help. The proposal to enable using a plugin system sounds very promising to us. We will do our best to help in every way possible to facilitate quick integration between our platform and the excellent tools you all have developed at sigstore. Do you think it will be available later this quarter? We want to estimate when we can enable our customers to work with your tool through our platform.

> Hey, thanks for the PR, and @akljph, sorry for not writing back via email.
> 
> The concern with taking on additional platforms is that the maintainers don't have familiarity with the provider, so as issues arise, debugging won't be trivial. At the same time, I would like to find a solution to unblock you and other new providers.
> 
> After chatting with maintainers, one thought we had was using a plugin system like https://github.com/hashicorp/go-plugin/blob/main/README.md. I plan to spend some time this week to see if it would be a good fit. At a glance, it seems quite promising - we create an interface for signing and verification that each provider would define, and then could load in plugins dynamically rather than bundling them in the package. If you have other suggestions too, I'd be happy to chat more.

haydentherapper:
We've completed https://github.com/sigstore/sigstore/issues/1658 to offer a plugin interface for KMS providers. Organizations can independently and privately develop & distribute their plugins without needing downstream updates to libraries to support additional KMS providers as build-time dependencies

See https://github.com/sigstore/sigstore/tree/main/pkg/signature/kms/cliplugin for more information and https://github.com/sigstore/sigstore/tree/main/test/cliplugin/localkms for an example.

Closing PR, if you have any questions, feel free to ping us. If you move forward with this implementation, feel free to ping us if you'd like your implementation to be highlighted on our upcoming blog post on this feature.

PR #1686, opened 2024-03-28
add support for verifying IEEE P1363 encoded ECDSA sigs

### LINKED ISSUE #1685: fail to verify IEEE P1363 encoded ECDSA signatures
**Description**

ECDSA signatures are often DER encoded, but there does exist an alternate standard encoding specified in IEEE P1363. Verifying a signature encoded in IEEE P1363 format fails

**Version**

v1.8.2

bobcallaway:
fyi @laurentsimon 

Closes #1685

### DIFF
--- a/pkg/signature/ecdsa.go
+++ b/pkg/signature/ecdsa.go
@@ -192,7 +192,9 @@ func (e ECDSAVerifier) VerifySignature(signature, message io.Reader, opts ...Ver
 		return fmt.Errorf("invalid ECDSA public key for %s", e.publicKey.Params().Name)
 	}
 
-	var asnParseTest []*big.Int
+	asnParseTest := struct {
+		R, S *big.Int
+	}{}
 	if _, err := asn1.Unmarshal(sigBytes, &asnParseTest); err == nil {
 		if !ecdsa.VerifyASN1(e.publicKey, digest, sigBytes) {
 			return errors.New("invalid signature when validating ASN.1 encoded signature")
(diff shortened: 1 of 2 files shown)


pkg/signature/ecdsa.go
```diff
@@ -190,8 +192,21 @@ func (e ECDSAVerifier) VerifySignature(signature, message io.Reader, opts ...Ver
 		return fmt.Errorf("invalid ECDSA public key for %s", e.publicKey.Params().Name)
 	}
 
-	if !ecdsa.VerifyASN1(e.publicKey, digest, sigBytes) {
-		return errors.New("invalid signature when validating ASN.1 encoded signature")
+	var asnParseTest []*big.Int
+	if _, err := asn1.Unmarshal(sigBytes, &asnParseTest); err == nil {
```
haydentherapper, line 196: Is this the correct check? I would have thought that a sequence needs to be represented as a struct?

haydentherapper: Maybe it's just flexible enough since it is effectively an array. Seems a little brittle, but if the tests pass, this should be sufficient.

haydentherapper (approved):

pkg/signature/ecdsa.go
bobcallaway: This is what was recommended by https://pkg.go.dev/encoding/asn1#Unmarshal

haydentherapper (approved):

PR #1728, opened 2024-05-09
Fix RSA signature ASN.1 parsing bug

malancas:
#### Summary

The Azure signer was encoding both ECDSA and RSA signatures with ASN.1. The RSA verifier in Cosign does not expect RSA signatures to be ASN.1 encoded, so the signatures returned by the Azure signer would fail verification against Cosign when used with a local public key.


Hayden-IO:
Going to close this out since it's been a bit, feel free to reopen if this is still an issue. We should probably move towards deprecating RSA across client tooling in favor of ECC.

Issue #1735, opened 2024-05-26
Hashicorp Vault KMS provider uses the wrong signature algorithm


berkitamas:
**Description**

I am building an on-premise sigstore stack using Hashicorp Vault as KMS backend. I noticed that `cosign sign` refuses to sign my image with the following error:
```
Error: signing [<image>]: getting signer: getting key from Fulcio: retrieving cert: POST https://fulcio.example.com/api/v1/signingCert returned 400 Bad Request: "{\"code\":3, \"message\":\"x509: signature over certificate returned by signer is invalid: crypto/rsa: verification error\", \"details\":[]}"
main.go:74: error during command execution: signing [<image>]: getting signer: getting key from Fulcio: retrieving cert: POST https://fulcio.example.com/api/v1/signingCert returned 400 Bad Request: "{\"code\":3, \"message\":\"x509: signature over certificate returned by signer is invalid: crypto/rsa: verification error\", \"details\":[]}"
```

Also `openssl ts verify` fails too with this error:
```
$ openssl ts -verify -in response.tsr -data "myblob" -CAfile root.pem -untrusted chain.pem
Using configuration from /opt/homebrew/etc/openssl@3/openssl.cnf
Verification: FAILED
008C61F901000000:error:0200008A:rsa routines:RSA_padding_check_PKCS1_type_1:invalid padding:crypto/rsa/rsa_pk1.c:79:
008C61F901000000:error:02000072:rsa routines:rsa_ossl_public_decrypt:padding check failed:crypto/rsa/rsa_ossl.c:746:
008C61F901000000:error:1C880004:Provider routines:rsa_verify:RSA lib:providers/implementations/signature/rsa_sig.c:785:
008C61F901000000:error:10800069:PKCS7 routines:PKCS7_signatureVerify:signature failure:crypto/pkcs7/pk7_doit.c:1125:
008C61F901000000:error:1780006D:time stamp routines:TS_RESP_verify_signature:signature failure:crypto/ts/ts_rsp_verify.c:148:
```

This is all because Vault uses PSS by default which is not recognized by cosign.

By adding `"signature_algorithm": "pkcs1v15"` to the signing request towards Vault, I could make both timestamp authority and cosign work.

**Version**
* cosign: 2.2.4
* Rekor: v1.3.6
* Fulcio: v1.4.5
* Timestamp Authority: v1.2.2

PR #1736, opened 2024-05-26
Fix Hashicorp Vault KMS to use PKCS1 v1.5

berkitamas:
#### Summary

By default, Hashicorp Vault uses pss for signing. Several (if not all) sigstore components want the signature in PKCS1 v1.5 format, and since this value is not set, they fail with verification error.

Resolves: #1735


haydentherapper (approved):
Thanks!

cpanato (approved):

PR #1753, opened 2024-06-14
Update to latest letsencrypt/boulder.

kommendorkapten:
#### Summary

This lets us use the `GoodKey` function to verify some properties of EC public keys too, which was ignored earlier.
This commit also bumps the go version as 1.22 is requried for this boulder version.

#### Release Note
* Updated go version to 1.22
* Updated to latest letsencrypt/boulder.
* Added support for verifying EC public keys in `cryptoutils/ValidatePubKey` function.

#### Documentation
N/A


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore
 
-go 1.21
+go 1.22.0
```
cpanato, line 3:
```suggestion
go 1.22.4
```
can we bump to that? of just 1.22

haydentherapper: IIRC you can't do `1.22` anymore without also adding `toolchain`, or you need to specify the specific go version like here.

haydentherapper (approved):
Thank you!

cpanato (approved):

PR #1761, opened 2024-06-20
Bump goodkey, fix breakage

jonjohnsonjr:
#### Summary

This fixes a breaking API change upstream:
https://github.com/letsencrypt/boulder/pull/7543

I believe the behavior is equivalent, but I am not a cryptography expert, so please let me know if this isn't quite right.

#### Release Note

NONE


hectorj2f (approved):

jonjohnsonjr:
Forgot to run `go mod tidy`, PTAL.

pkg/cryptoutils/publickey.go
```diff
@@ -148,8 +148,14 @@ func ValidatePubKey(pub crypto.PublicKey) error {
 	//   * Public key Q's x and y are within [0, p-1]
 	//   * Public key Q is on the curve
 	//   * Public key Q's order matches the subgroups (nQ = Ø)
-	p, err := goodkey.NewKeyPolicy(&goodkey.Config{FermatRounds: 100}, nil)
-	p.AllowECDSANISTP521 = true
+	allowedKeys := goodkey.LetsEncryptCPS()
```
haydentherapper, line 151: In case their definition of an allowed key ever changes, can we set each of the key fields to true? Also helps with readability.

haydentherapper (commented):
LGTM, just one small change

pkg/cryptoutils/publickey.go
jonjohnsonjr: Like so?

haydentherapper (dismissed):
Thanks! We will want to cut another release with this fix.

cpanato:
@jonjohnsonjr i've update the pr to sync the go mod to the kms part

cpanato (approved):

haydentherapper (approved):

jonjohnsonjr:
I don't mind at all, thanks!

PR #1794, opened 2024-08-03
Support email_verified as a String

sabre1041:
#### Summary

Added support for the `email_verified` property of an ID token to be either a boolean or string

Related to: https://github.com/sigstore/fulcio/pull/1744

#### Release Note

Added support for the `email_verified` property of the ID token to be either a String or Boolean

-->

#### Documentation

Added support for the `email_verified` property of the ID token to be either a String or Boolean

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Docker Build
         working-directory: ./test/e2e
-        run: docker-compose build
+        run: docker compose build
 
       - name: Integration test
         run: make test-e2e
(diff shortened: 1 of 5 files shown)


haydentherapper (commented):
Can you update the e2e test as well? - https://github.com/sigstore/sigstore/blob/89b95852bc1b8e05046bafce7d61871c341d8282/pkg/oauth/oidc/interactive_e2e_test.go#L35

haydentherapper:
The error is likely because `docker-compose` is now `docker compose`, can you switch those out in the e2e tests?

haydentherapper (approved):

PR #1809, opened 2024-08-15
fuzzing: improve coverage

AdamKorcz:
#### Summary

Adds coverage to Sigstore fuzzing by adding new fuzzers that target APIs that are not currently covered. 

I have added most of the tests in `test/fuzz` which holds all the existing fuzz tests, and I have added a single one in the directory of the API it tests because it targets an internal API.


test/fuzz/pem/fuzzcert_test.go
```diff
@@ -84,3 +86,94 @@ func FuzzUnmarshalPEMToPublicKey(f *testing.F) {
...
+		if err != nil {
+			return
+		}
+
+		cryptoutils.MarshalPrivateKeyToPEM(privateKey)
+		cryptoutils.MarshalPrivateKeyToEncryptedDER(privateKey, passFunc)
+	})
+}
+
+func FuzzParseCSR(f *testing.F) {
```
haydentherapper, line 126: I'm iffy on if we need this one, because it's just a call to `pem.Decode` and `x509.ParseCertificateRequest` with some conditionals. I suspect we'll mostly see timeouts on this fuzzer, so I'd probably remove.

haydentherapper (commented):
Thanks! Just need to fix linters

haydentherapper (approved):

PR #1811, opened 2024-08-17
Deserialize ed25519 keys from hashivault correctly

### LINKED ISSUE #1810: Hashivault ED25519 sign/verify fails
**Description**

Tried using an ed25519 key in a hashicorp vault transit engine with cosign. Failed with "PEM decoding failed" error.

Vault returns an ed25519 public key as the base64 encoding of the raw public key bytes with no further encoding (so no PEM to decode).

**Version**

cosign v2.4.0

stevenjohnstone:
#### Summary

Fixes sign/verify using an ed25519 key stored in Hashicorp Vault's transit engine. Vault returns public keys in different encodings depending on the key type. In the case of ed25519, it base64 encodes the raw public key bytes.

Resolves #1810

### DIFF
--- a/pkg/signature/kms/hashivault/client.go
+++ b/pkg/signature/kms/hashivault/client.go
@@ -231,13 +231,15 @@ func (h *hashivaultClient) fetchPublicKey(_ context.Context) (crypto.PublicKey, 
 		return nil, fmt.Errorf("could not parse public key pem as string")
 	}
 	// vault returns the key type in the "name" field
-	keyType := keyMap["name"]
-	if keyType == "ed25519" {
+	if keyType := keyMap["name"]; keyType == "ed25519" {
 		// vault returns ed25519 public keys as base64 encoding of
 		// the raw bytes of the key
 		decodedPublicKey, err := base64.StdEncoding.DecodeString(strPublicKey)
 		if err != nil {
 			return nil, fmt.Errorf("failed to base64 decode ed25519 public key: %s", err)
+		}
+		if keyLen := len(decodedPublicKey); keyLen != ed25519.PublicKeySize {
+			return nil, fmt.Errorf("decoded ed25519 public key length is %d, should be %d", keyLen, ed25519.PublicKeySize)
 		}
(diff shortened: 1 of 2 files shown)


pkg/signature/kms/hashivault/client.go
```diff
@@ -220,17 +221,28 @@ func (h *hashivaultClient) fetchPublicKey(_ context.Context) (crypto.PublicKey,
...
 	}
 
-	strPublicKeyPem, ok := publicKeyPem.(string)
+	strPublicKey, ok := publicKey.(string)
 	if !ok {
 		return nil, fmt.Errorf("could not parse public key pem as string")
 	}
+	// vault returns the key type in the "name" field
+	keyType := keyMap["name"]
+	if keyType == "ed25519" {
```
bobcallaway, line 235:
```suggestion
	if keyType := keyMap["name"]; keyType == "ed25519" {
```

pkg/signature/kms/hashivault/client.go
```diff
@@ -220,17 +221,28 @@ func (h *hashivaultClient) fetchPublicKey(_ context.Context) (crypto.PublicKey,
...
+		// vault returns ed25519 public keys as base64 encoding of
+		// the raw bytes of the key
+		decodedPublicKey, err := base64.StdEncoding.DecodeString(strPublicKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to base64 decode ed25519 public key: %s", err)
+		}
```
bobcallaway, line 241: perhaps worth adding a defensive check to ensure `len(decodedPublicKey) == ed25519.PublicKeySize`?

pkg/signature/kms/hashivault/client.go
stevenjohnstone: Changed, thanks :+1:

pkg/signature/kms/hashivault/client.go
stevenjohnstone: Yes, agree it's worth adding a check here. Thanks!

bobcallaway (approved):

PR #1866, opened 2024-10-16
Add workflow to synchronize go dependencies in submodules

codysoyland:
Signed-off-by: Cody Soyland <codysoyland@github.com>

#### Summary

This PR attempts to improve the workflow of managing Dependabot updates to this repo. The problem is that when Dependabot updates a dependency in the main module, those updates are not reflected in the "submodules" (kms providers) which includes a reference to the main module. That causes the build to fail, and one maintainer must manually run go mod tidy in the submodule and push those changes, and another maintainer must approve and merge them (as the last committer is disallowed from merging a PR). 

This PR depends on a new secret to be created `GO_MOD_SYNC_TOKEN`, which should have permission to write to the contents of this repo.

This is similar to the workflow in this PR: https://github.com/sigstore/sigstore-go/pull/314

### DIFF
--- a/.github/workflows/sync-go-mod.yml
+++ b/.github/workflows/sync-go-mod.yml
@@ -1,0 +1,49 @@
+name: Synchronize go.mod in submodules
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - opened
+      - synchronize
+    paths:
+      - 'go.mod'
+
+jobs:
+  sync-go-mod:
+    if: github.actor == 'dependabot[bot]'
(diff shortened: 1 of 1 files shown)


haydentherapper:
@codysoyland can this be replaced by the dependabot changes you made in sigstore-go?

codysoyland:
> @codysoyland can this be replaced by the dependabot changes you made in sigstore-go?

Possibly...  it looks like this repo is already using Dependabot grouped updates, so we can probably just close this. 👍🏻

PR #1865, opened 2024-10-15
bump to go 1.22.8

cpanato:
- bump to go 1.22.8

PR #1878, opened 2024-11-05
fix: set go module directive to 1.22.0

### LINKED ISSUE #1879: go directive in go.mod is unnecessarily specific
**Description**

The go directive sets the minimum version of Go required to use this module. There's no need to set this to the latest semver patch version of a given Go release unless the semantics of that version of Go are required to build/test/use the module.

Notably since PR https://github.com/sigstore/sigstore/pull/1865 any consumers of sigstore as a library would need to bump their own go directive to 1.22.8 in order to pickup the sigstore 1.8.10 release.

dnwe:
#### Summary

The go directive sets the minimum version of Go required to use this module. There's no need to set this to the latest semver patch version of a given Go release unless the semantics of that version of Go are required to build/test/use the module.

Notably since PR #1865 any consumers of sigstore as a library would need to bump their own go directive to 1.22.8 in order to pickup the sigstore 1.8.10 release.

Ref: https://go.dev/ref/mod#go-mod-file-go

Fixes #1879

#### Release Note

None

#### Documentation

N/A

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -16,24 +16,37 @@
 name: CodeQL
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ main ]
+    branches: [main]
 
 permissions:
   contents: read
   security-events: write
 
 jobs:
+  go_version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+      - name: Get Go version directive from go.mod as major.minor
+        id: list
+        run: |
+          go list -m -f 'version={{.GoVersion}}' | cut -d. -f-2 >>"${GITHUB_OUTPUT}"
+    outputs:
+      version: ${{ steps.list.outputs.version }}
   analyze:
     name: Analyze
+    needs:
+      - go_version
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'go' ]
+        language: ["go"]
     steps:
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -41,7 +54,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
-          go-version-file: go.mod
+          go-version: "${{ needs.go_version.outputs.version }}.x"
           check-latest: true
 
       # Initializes the CodeQL tools for scanning.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -26,35 +26,25 @@ permissions:
   security-events: write
 
 jobs:
-  go_version:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
-      - name: Get Go version directive from go.mod as major.minor
-        id: list
-        run: |
-          go list -m -f 'version={{.GoVersion}}' | cut -d. -f-2 >>"${GITHUB_OUTPUT}"
-    outputs:
-      version: ${{ steps.list.outputs.version }}
   analyze:
     name: Analyze
-    needs:
-      - go_version
     runs-on: ubuntu-latest
-
     strategy:
       fail-fast: false
       matrix:
         language: ["go"]
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Get Go version directive from go.mod as major.minor
+        id: go_version
+        run: |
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore
 
-go 1.22.8
+go 1.22.0
```
cpanato, line 3: we will need to update places like https://github.com/sigstore/sigstore/blob/main/.github/workflows/e2e_test.yml#L37C1-L37C34 to pin to `1.22` or `1.23` otherwise it will install the 1.22.0 only and i think we want to use the latest available always

cpanato: cc @haydentherapper

dnwe: Good spot, pushed up [d8cb131](https://github.com/sigstore/sigstore/pull/1878/commits/d8cb131efa0f6cb70b4f8f192adcd89c2279b216) as a suggestion of how to address that

haydentherapper: Caveat: I'm not an expert on Go version resolution

The reason we began to include the patch release version is because the tooling required us to either set the specific version of Go or add a `toolchain` directive. When @cpanato started moving us to 1.23, it seemed like something changed in the tooling and we didn't need to specify the patch version, only the minor version like what we used to do.

If it is possible to only specify `1.22` or `1.23`, let's do that. I'd rather not have the complexity of CI tooling pulling a different version than what's in the go.mod file.

dnwe: Certainly you can't use `go 1.22` in go.mod, in recent go1.22.x versions they'll just complain that go.mod needs tidying and will update it to `1.22.0` and also add a `toolchain go1.22.8` directive:
```
% grep -E '^(go|toolchain)' go.mod
go 1.22
% go1.22.8 build ./...
go: updates to go.mod needed; to update it:
        go mod tidy
% go1.22.8 mod tidy
% grep -E '^(go|toolchain)' go.mod
go 1.22.0
toolchain go1.22.8
```

dnwe: Something does seem to have changed in 1.23 whereby '1.23' is seemingly accepted as a directive again

```
% grep -E '^(go|toolchain)' go.mod
go 1.23
% go1.23.2 build ./...
% go1.23.2 mod tidy
% grep -E '^(go|toolchain)' go.mod
go 1.23
```

dnwe: The change in behaviour in 1.23 is referenced in this comment on this well-cited GH issue on the confusion around the go directive changes:

image

https://github.com/golang/go/issues/62278#issuecomment-2062002018

However, that's really saying that `go` itself will just silently assume 1.23.0 for module requirements and toolchain rather than the more useful 1.23.X. 

Unfortunately from a GitHub Actions perspective, setup-go doesn't currently know or use go toolchain versions, it's just manually parsing the go.mod for 'go' directives itself and hasn't really kept up with the 1.22 and 1.23 changes here. There's an issue tracking that over [here](https://github.com/actions/setup-go/issues/457)

cpanato: thank you @dnwe for all the clarification TIL a lot about this fuzzy thing that i cant understand

.github/workflows/codeql.yml
```diff
@@ -16,32 +16,45 @@
...
+        language: ["go"]
     steps:
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Set up Go
         uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
         with:
-          go-version-file: go.mod
+          go-version: "${{ needs.go_version.outputs.version }}.x"
```
cpanato, line 57: in the action, we don't need to set the .x, only major.minor is fine

.github/workflows/codeql.yml
```diff
@@ -16,32 +16,45 @@
...
-    branches: [ main ]
+    branches: [main]
 
 permissions:
   contents: read
   security-events: write
 
 jobs:
+  go_version:
+    runs-on: ubuntu-latest
```
cpanato, line 30: i dont think we need a exta job just to extract the version we can do in the same job

cpanato (changes requested):
some small comments

.github/workflows/codeql.yml
dnwe: My understanding from setup-go is that `1.22` is analogous to `1.22.0` whereas `1.22.x` is analogous to "use latest patch version" and so (as of today) would get `1.22.9` which is what we desire for testing?

.github/workflows/codeql.yml
dnwe: Sure we can in-line it. The only reason I get it separate was that I was hoping it would be throwaway code that could eventually be removed once setup-go handles all of this better :)

.github/workflows/codeql.yml
cpanato: with the check-latest it will pick up the latest one, but i am okay to keep the .x not a big thing

.github/workflows/codeql.yml
cpanato: thanks

.github/workflows/codeql.yml
dnwe: ah yes you're right, gave it a quick test:

```
Run actions/setup-go@v5
  with:
    go-version: 1.22
    check-latest: true
    token: ***
    cache: true
Setup go version spec 1.22
Attempting to resolve the latest version from the manifest...
matching 1.22...
Resolved as '1.22.9'
Attempting to download 1.22.9...
matching 1.22.9...
Acquiring 1.22.9 from https://github.com/actions/go-versions/releases/download/1.22.9-11715914133/go-1.22.9-linux-x64.tar.gz
```

.github/workflows/codeql.yml
dnwe: ah wait, I remember why I made it a separate task now — chicken before the egg

I needed Go installed in order to use `go list -m -f 'version={{.GoVersion}}'`, but I can switch to just using awk and doing `awk -F'[ .]' '/^go /{print $2 "." $3; exit}' go.mod`

dnwe: Updated to in-line the steps into the existing job and to use awk

cpanato (approved):
lgtm


thanks

@haydentherapper your final review

go.mod
haydentherapper: Thanks for digging into this!

haydentherapper (approved):

PR #1888, opened 2024-11-12
Replace custom auth code with `azidentity.NewDefaultCredential` for Azure KMS client

malancas:
#### Summary

This replaces custom authentication code used in the Azure KMS client with `azidentity.NewDefaultCredential`, which will try each type of authentication method.


cpanato (approved):

cpanato (approved):
cool thanks

Issue #1827, opened 2024-09-07
Support Alibaba Cloud KMS


mozillazg:
**Description**

Support Alibaba Cloud KMS.

haydentherapper:
We've completed https://github.com/sigstore/sigstore/issues/1658 to offer a plugin interface for KMS providers. Organizations can independently and privately develop & distribute their plugins without needing downstream updates to libraries to support additional KMS providers as build-time dependencies

See https://github.com/sigstore/sigstore/tree/main/pkg/signature/kms/cliplugin for more information and https://github.com/sigstore/sigstore/tree/main/test/cliplugin/localkms for an example.

PR #1901, opened 2024-12-06
draft: add plugin support for KMSs 

### LINKED ISSUE #1658: Add plugin interface for out-of-tree KMS providers
**Description**

Investigate using [go-plugin](https://github.com/hashicorp/go-plugin/blob/main/README.md) as an approach for out-of-tree KMS providers to integrate with Sigstore tooling without merging code into this repo.

I believe this is the correct interface to base the plugin design off of: https://github.com/sigstore/sigstore/blob/8a49902a31ccd8f4d297d8109a17e9bb78f8a1ba/pkg/signature/kms/kms.go#L72

ramonpetgrave64:
Resolves #1658 

#### Summary

[TODO: Spec]()

We're adding plugin support the sigstore to support separately-compiled KMS providers. We experimented with the hashicorp/go-plugin system suggested in #1658, and we would like to try kubectl's style of supporting plugin programs. It basically works like this.

- The plugin is a separate program on the system's PATH, invoked as a command
- the `SignerVerifier` interface arguments are sent as CLI arguments and over stdin
- the `SignerVerifier` interface return values are sent over stdout

You should be able to implement a plugin in any language, even shellscript. If using go, once you implement the `SignerVerifier` interface methods, your plugin program could be as simple as the example included in this PR at "test/cliplugin/localkms/main.go".

#### Release Note

TODO: ...

#### Documentation

TODO: ...

### Testing

To test this PR, also clone the code from another draft PR to sigstore/cosign #xxxx.

#### local KMS plugin

Use the test local KMS plugin from this PR

1. compile cosign and the plugin

    ```
    go build -C cosign/cmd/cosign -o `pwd`/cosign-cli
    go build -C sigstore/test/cliplugin/localkms -o `pwd`/sigstore-kms-localkms
    ```

2. generate an RSA key

    ```
    cosign-cli generate-key-pair --kms sigstore-kms-localkms://`pwd`/key.pem
    ```

3. sign some data

    ```
    echo hello > data.txt
    cosign-cli sign-blob data.txt --tlog-upload=false --key sigstore-kms-localkms://`pwd`/key.pem
    openssl dgst -sha256 -sign key.pem data.txt | base64
    ```

## TODO

- [ ] link to spec
- [ ] add docstrings in code
- [ ] add markdown documentation
- [ ] complete the remaining `SignerVerifier` interface methods
- [ ] add more unit tests
- [ ] show example plugin that reuses existing GCP KMS implementation as a plugin
- [ ] test with Alibaba KMS #1827 
- [ ] add example reusable workflow for plugin authors to have CI tests


pkg/signature/kms/cliplugin/signer.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// TODO: Additonal methods to be implemented
+
+func (c PluginClient) DefaultAlgorithm() string {
+	args := &common.MethodArgs{
+		MethodName:       common.DefaultAlgorithmMethodName,
+		DefaultAlgorithm: &common.DefaultAlgorithmArgs{},
+	}
+	resp, err := c.invokePlugin(context.TODO(), nil, args)
```
haydentherapper, line 101: Can we use the context that comes from `AddProvider`?

.github/workflows/test.yml
```diff
@@ -70,3 +79,60 @@ jobs:
...
+      # TODO: remove my personal forks' references.
+      COSIGN_REPO: ramonpetgrave64/cosign
+      COSIGN_REF: ramonpetgrave64-kms-plugin-cli
+
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: sigstore
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
haydentherapper, line 105: Why do we need Cosign for these tests? Could we instead have a test that imports cliplugin and calls all of the implemented functions, `CreateKey`, etc? This would be more lightweight and easy to iterate on.

haydentherapper (commented):
This is looking good!

.github/workflows/test.yml
ramonpetgrave64: We could do that, but we would have to trigger that unit test with a special build flag, since the invoked of the unit test would first need to compile a real plugin binary first, probably our localkms implementation, since it also uses our Dispatch functions. And then the workflow could trigger this unit test with the special build flag. We also planned to make this build flag configurable, so plugin authors could run sigstore unit tests against their own prebuilt plugin programs. 

I like to know we’re still compatible with cosign, so I think we should do both.

test/cliplugin/localkms/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/handler"
+)
+
+const expectedProtocolVersion = "1"
+
+func main() {
+	// we log to stderr, not stdout. stdout is reserved for the plugin return value.
+	spew.Fdump(os.Stderr, os.Args)
```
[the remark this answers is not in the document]
ramonpetgrave64: Yes, because most of the strict fields are pointers within nested structs. fmt will not resolve the values at pointers beyond the first level, but spew will.

test/cliplugin/localkms/main.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	resp, err := handler.Dispatch(os.Stdout, os.Stdin, pluginArgs, signerVerifier)
+	if err != nil {
+		log.Fatal(err)
+	}
+	spew.Fdump(os.Stderr, resp)
+}
+
+// don't do this because it's more clean if the host redirects the child's stderr to its own stderr
```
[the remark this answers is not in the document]
ramonpetgrave64: Yes, I’ll add a section to the readme about error handling.

.github/workflows/test.yml
ramonpetgrave64: Added in the in `signer_program_test.go`, Makefile, `test.yml`, and `e2e_test.yml`. We won't need to run it on both workflows, but I'm not sure where is more appropriate. wdyt?

test/cliplugin/localkms/go.mod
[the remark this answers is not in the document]
ramonpetgrave64: added.

.github/workflows/test.yml
haydentherapper: e2e_test tests the cloud provider integrations. That seems like a good place to have this test as well. Putting these tests besides build tags SGTM.

I would omit Cosign testing here, as this is a general-purpose library that doesn't need to be aware of its dependents.

ramonpetgrave64: Alright, removed.

pkg/signature/kms/kms.go
```diff
@@ -50,11 +55,25 @@ var providersMap = map[string]ProviderInit{}
...
-			return pi(ctx, keyResourceID, hashFunc, opts...)
+			sv, err := pi(ctx, keyResourceID, hashFunc, opts...)
+			if err != nil {
+				return nil, err
+			}
+			return sv, nil
+		}
+	}
+	if pi, ok := providersMap[CLIPluginProviderKey]; ok {
+		sv, err := pi(ctx, keyResourceID, hashFunc, opts...)
```
jku, line 72: Is there a reason to limit the system to one registered plugin at a time? This seems like a change compared to current setup (assuming we at some point follow the plan of implementing current KMS providers as plugins).

At first glance I don't see why CLIPluginProviderKey exists -- could the plugins use the plugin name as the argument to AddProvider, and then this code could do the same thing the existing provider matching does: if the keyResourceID has a prefix that matches the provider key, use that provider?

jku (commented):
I'm probably not qualified to approve this code but some questions on the design:
* It looks like exactly one plugin can be registered at a time... Maybe this is as designed but I did not understand that from the design doc so left a comment in code
* I think the approach in the examples and PR description is a bit problematic: `generate-key-pair` is used as the core example when that is the part that real KMS plugins will likely never even implement
  * Signing is the part that IMO should be shown (as it shows that the plugin has to be able to identify the private key using the keyref)
   * maybe even your localkms plugin should not handle private key generation: it could even just use a hard coded private key to make it as simple as possible (meaning the private key is still loaded like it is now but stored as a file in git)

pkg/signature/kms/kms.go
```diff
@@ -50,11 +55,25 @@ var providersMap = map[string]ProviderInit{}
 // Get returns a KMS SignerVerifier for the given resource string and hash function.
 // If no matching provider is found, Get returns a ProviderNotFoundError. It
 // also returns an error if initializing the SignerVerifier fails.
+// If keyResourceID doesn't match any of our hard-coded provers' schemas,
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

pkg/signature/kms/kms.go
ramonpetgrave64: You can actually have many plugin programs on the system, and you can choose which plugin program based on the `keyResourceID`: e.g., `my-hsm:///my-key` or `digitalocean]://my-key`. 

What's "registered" is actually only the provider's initialization function that produces a `SignerVerifier` implementation. `CLIPluginProviderKey` only exists because we need a way to access `clipugin`'s the initialization function, but we cannot import it from the `kms` package because that would cause a circular import, since `cliplugin` already imports `kms`. So, when `cliplugin` is imprted, its package `init()` will call `AddProvider(CLIPluginProviderKey, signer_verifier_maker_func)`.

> could the plugins use the plugin name as the argument to AddProvider

Yes, but that would mean that the main program has to first discover all potential plugin programs, and then register each of their initialization functions. I thought it simpler to just first determine of the caller intended to use a plugin or not. Then, determining which plugin to use is only a matter of parsing the keyResourceID and finding the corresponding program on the system.

Eventually when all the providers become plugins, we won't need this registration process. But for now, we need to keep the diff small and not break compatibility.

jku: > You can actually have many plugin programs on the system

do you mean you can have multiple binaries installed on the system? That seems correct, but a single application can only load one plugin, correct? Or more precisely, only the last loaded plugin is actually used. 

Now this may be as designed but 
* I didn't understand this from the design doc so wanted to make sure
* this seems like a change compared to current KMS implementation: I think you can currently use multiple providers (unsure if anyone is interested in this)

ramonpetgrave64:
@jku I agree `SignerVerifier.CreateKey()` maybe shouldn't be the core example, but for now it's the simplest to show to keep the PR small, because it doesn't need all the optional arguments that other methods like `SignerVerififer.SignMessage()` need. 

If folks think I should go ahead with implementing another method for this PR, I'm open to it.

As for the example implementation, I don't think it's too unusual to want to use a KMS to create new ephemeral signing keys. https://docs.sigstore.dev/cosign/signing/overview/#verifying-identity-and-signing-the-artifact

pkg/signature/kms/kms.go
ramonpetgrave64: I can test this later, but: A single application should be be able to load many plugins simultaneously via `Get()`. The returned `PluginClient` doesn't modify any global variables, so each instance should be independent.

As for many instances of the same PluginClient that all talk to the same plugin program, it would be up to the authors to make their plugins be able to handle multiple simultaneous CLI invocations (like TUF caching).

jku:
> I don't think it's too unusual to want to use a KMS to create new ephemeral signing keys

:eyes: huh, do people use KMS keys as **ephemeral** signing keys? This sounds kind of wild to me but I'll take your word for it.

How does the "ephemeral" part work? Who deletes the keys?

ramonpetgrave64:
> How does the "ephemeral" part work? Who deletes the keys?

AWS has a `ScheduleKeyDeletion` API. I don't know about other's however. Still, it's up to the author or user.

jku:
> AWS has a `ScheduleKeyDeletion` API. I don't know about other's however. Still, it's up to the author or user.

My question is, is this actually used for sigstore ephemeral keys? Ephemeral keys on a KMS sounds like a fundamentally _bad idea_ to me (insecure, inefficient and complicated all at once) and we should not make it possible unless this is really an important cosign feature...  

Also we absolutely cannot leave that to the user to handle: if plugins are meant to be used for ephemeral signing then plugins absolutely must handle key deletion. But I reiterate that supporting this sounds really wrong to me.

pkg/signature/kms/kms.go
jku: Get() looks up the plugin provider with a constant key. I don't see how it could return different plugin providers

haydentherapper: `providersMap[CLIPluginProviderKey]` returns a function to load a SignerVerifier, and it isn't caching that result anywhere. If you invoke `Get()` repeatedly with different keys, I believe it should create multiple `SignerVerifier` clients.

ramonpetgrave64:
@jku Spoken offline, it's worth considering disabling that, and ideally `CreateKey` is not the first example for the plugin system, but technically, it's the easiest to demonstrate in a small PR. I can followup within the next few small PRs to have a complete `SignMessage` example.

In the meantime, I think we should followup this discussion in https://github.com/sigstore/sigstore/issues/1914

ramonpetgrave64:
Example GCP plugin here https://github.com/ramonpetgrave64/sigstore-kms-mygcpkms

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+`
+
+### Protocol
+
+The main program will invoke the program with these specifications:
+
+* stdin
+  * Data to be signed or verified.
+* arg 1
+  * A number idnetifying the version of this protocol.
```
haydentherapper, line 22:
```suggestion
  * A number identifying the version of this protocol.
```

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+    * In the future we may change this protocol, either the encoding or the formatting of argument and return values.
+    These are necessarily breaking changes, and will be major-version bumps to sigstore and cosign.
```
haydentherapper, line 24:
```suggestion
    Protocol changes will result in a major-version bump for the library.
```

haydentherapper: Removing a mention of Cosign, as the maintainers may not want to do a major bump for this.

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Error Handling
+
+The plugin program’s stderr will be redirected to the main program’s stderr. This way, the main program may also see the plugin program’s debug messages.
+
+Plugin authors may return errors with `PluginResp.ErrorMessage`, but the plugin's exit status will be ignored.
+
+### Implementation
+
+Plugin authors must implement the `kms.SignerVerifier` interface methods in their chosen language. Each method will inbvoke your program once, and the reponse will be parsed from stdout.
```
haydentherapper, line 45:
```suggestion
Plugin authors must implement the `kms.SignerVerifier` interface methods in their chosen language. Each method will invoke your program once, and the response will be parsed from stdout.
```

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+Exit status is ignored. Your programs stderr will be redirected to the main program, and errors you wish to return must be serialized in `PluginResp.ErrorMessage` in stdout.
```
haydentherapper, line 47:
```suggestion
Exit status is ignored. Your program's stderr will be redirected to the main program, and errors you wish to return must be serialized in `PluginResp.ErrorMessage` in stdout.
```

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+For authors using Go, we vend some helper functions to help you get started.
+
+* `GetPluginArgs(osArgs []string) (*PluginArgs, error)`
+  * to parse the `PluginArgs`. The plugin author can use the values for some pre-processing, perhaps with the values in `PluginArgs.InitOptions`.
```
haydentherapper, line 52: tiny nit, i'd make each bullet point one line rather than have single sub bullet points

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+* `WriteErrorResponse(wr io.Writer, err error) error`
+  * , should the author simply send an error message and exit.
+
+## Development
+
+Changes to the `SignerVerifier` interface are to be handled in [./signer.go's](./signer.go) `PluginClient` and [./handler/dispatch.go's](./handler/dispatch.go) `Dispatch()`.
+
+### Adding New Methods or Method Options
+
+Adding new methods or options are *not* necessarily breaking changes to the schemas, so we may consider these to be minor version increments, both to the protocol version and the sigstore version.
```
haydentherapper, line 66: When would adding a method be a breaking change?

pkg/signature/kms/cliplugin/client.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	// PluginBinaryPrefix is the prefix for all plugin binaries. e.g., sigstore-kms-my-hsm.
+	PluginBinaryPrefix = "sigstore-kms-"
+)
+
+// init registers the plugin system as a provider. It does not search for plugin programs.
+// users must import this package, e.g.,  `import _ "github.com/sigstore/sigstore/pkg/signature/kms/cliplugin"`
+func init() {
+	kms.AddProvider(kms.CLIPluginProviderKey, func(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts ...signature.RPCOption) (kms.SignerVerifier, error) {
```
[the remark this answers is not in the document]
haydentherapper: I had this same thought about simplifying this client, but with this current approach, we can easily import this plugin provider into Cosign right now while retaining the other KMS providers. Maybe we go back and refactor this provider later on once we've migrated all other providers to their own plugins.

pkg/signature/kms/cliplugin/client.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/common"
+)
+
+const (
+	// PluginBinaryPrefix is the prefix for all plugin binaries. e.g., sigstore-kms-my-hsm.
+	PluginBinaryPrefix = "sigstore-kms-"
+)
+
+// init registers the plugin system as a provider. It does not search for plugin programs.
+// users must import this package, e.g.,  `import _ "github.com/sigstore/sigstore/pkg/signature/kms/cliplugin"`
```
haydentherapper, line 36:
```suggestion
// Users must import this package, e.g.,  `import _ "github.com/sigstore/sigstore/pkg/signature/kms/cliplugin"`
```

pkg/signature/kms/cliplugin/common/interface.go
```diff
@@ -0,0 +1,81 @@
...
+	HashFunc        crypto.Hash `json:"hashFunc"`
+	// TODO: extracted values from signature.RPCOption from LoadSignerVerifier().
+}
+
+// MethodArgs contains the method arguments. MethodName must be specified,
+// while any one of the other fields describing method arguments must also be specified.
+type MethodArgs struct {
+	// MethodName specifies which method is intended to be called.
+	MethodName       string                `json:"methodName"`
+	DefaultAlgorithm *DefaultAlgorithmArgs `json:"defaultAlgorithm,omitempty"`
```
haydentherapper, line 54: Given this returns a constant string for a given plugin, can we switch this to a field in InitOptions, and then `DefaultAlgorithm()` should just return that string?

pkg/signature/kms/cliplugin/common/interface.go
```diff
@@ -0,0 +1,81 @@
...
+
+type DefaultAlgorithmArgs struct {
+}
+
+type DefaultAlgorithmResp struct {
+	DefaultAlgorithm string
+}
+
+type CreateKeyArgs struct {
+	CtxDeadline *time.Time
```
haydentherapper, line 75: Could this come from InitOptions rather than having to define this as a parameter for every method?

pkg/signature/kms/cliplugin/handler/dispatch.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/common"
+)
+
+// GetPluginArgs parses the PluginArgs from the os args.
+func GetPluginArgs(osArgs []string) (*common.PluginArgs, error) {
```
haydentherapper, line 29: should check the length of `osArgs` before accessing the array

pkg/signature/kms/cliplugin/handler/dispatch.go
```diff
@@ -0,0 +1,74 @@
...
+	argsStr := osArgs[2]
+	var args common.PluginArgs
+	if err := json.Unmarshal([]byte(argsStr), &args); err != nil {
+		return nil, err
+	}
+	return &args, nil
+}
+
+// WriteResponse writes JSON-serialized PluginResp to the outoput.
```
haydentherapper, line 38:
```suggestion
// WriteResponse writes JSON-serialized PluginResp to the output.
```

pkg/signature/kms/cliplugin/signer.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	if resp.ErrorMessage != "" {
+		return nil, fmt.Errorf("%w: %s", ErrorPluginReturnError, resp.ErrorMessage)
+	}
+	return &resp, nil
+}
+
+// TODO: Additonal methods to be implemented
+
+func (c PluginClient) DefaultAlgorithm() string {
```
haydentherapper, line 96: As mentioned in the other comment, given this takes no arguments and returns a constant string, can this be an initialization param rather than requiring invoking the plugin?

pkg/signature/kms/cliplugin/signer.go
```diff
@@ -0,0 +1,124 @@
...
+		DefaultAlgorithm: &common.DefaultAlgorithmArgs{},
+	}
+	resp, err := c.invokePlugin(context.TODO(), nil, args)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return resp.DefaultAlgorithm.DefaultAlgorithm
+}
+
+func (c PluginClient) CreateKey(ctx context.Context, algorithm string) (crypto.PublicKey, error) {
```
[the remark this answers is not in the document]
haydentherapper: It'd be helpful to document the purpose of `handler`. If I'm understanding correctly, `handler` is helper functions for implementers in Go? Is there a reason why we need the indirection with the additional functions in handler/methods.go instead of `Dispatch` calling `pluginClient.Method()`?

pkg/signature/kms/cliplugin/signer_program_test.go
```diff
@@ -0,0 +1,146 @@
...
+		ProtocolVersion: common.ProtocolVersion,
+		KeyResourceID:   keyResourceID,
+		HashFunc:        testHashFunc,
+		// TODO: include extracted values from opts
+	}
+	pluginClient := newPluginClient(executable, initOptions)
+	return pluginClient
+}
+
+func Test_DefaultAlgorithm(t *testing.T) {
```
haydentherapper, line 61:
```suggestion
func TestDefaultAlgorithm(t *testing.T) {
```

haydentherapper: No need for the underscore in test names

test/cliplugin/localkms/README.md
```diff
@@ -0,0 +1,14 @@
+# Local KMS
+
+This module is an example implementation of a KMS plugin.
+See [clicplugin](../../../pkg/signature/kms/cliplugin/).
```
haydentherapper, line 4:
```suggestion
See [cliplugin](../../../pkg/signature/kms/cliplugin/).
```

test/cliplugin/localkms/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+The Key Ref is a path to am RSA private key on your system.
```
haydentherapper, line 6:
```suggestion
The Key Ref is a path to a RSA private key on your system.
```

test/cliplugin/localkms/implementation.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("error generating private key: %w", err)
+	}
+
+	privateKeyPEM := &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(privateKey),
+	}
+
+	path := i.keyResourceID
+	privateKeyFile, err := os.Create(path)
```
haydentherapper, line 68: To confirm, is the key being written to disk since it'll later be used by SignMessage? If so, worth a comment to mention you shouldn't do this in a real implementation.

goatwu1993:
Really excited of this feature. My experience is the GetPublicKey should be implemented first. Once it is done cosign main program would know the key algorithm, useful when signing. 

Our company's use case.

1. client: forked cosign
2. server: https server. required auth. Owned by company infrastructure team. backed by awskms. Expected authorization + full payload (base64 encoded)

If payload is always hashed the validation/auditing at Sign server will be troublesome. client would have to dump the payload to somewhere like file or env so plugin can read it and compare with the hash send from cosign main program before sending to the sign server.

My question is could cosign not hash (or at least have an option, if sending payload to the plugin is considered insecure) so the full payload can be sent to the plugin?

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,83 @@
...
+This is a package and module for using custom KMS plugins as separate executables. 
+It is intended to be used by cosign, but you may use this in your own programs that import sigstore.
+
+## Design
+
+We follow [kubectl's style](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#writing-kubectl-plugins) of plugins. Any language that can create CLI programs can be used to make a plugin.
+
+### Usage
+
+Plugins are separate programs on your system's PATH, named in the scheme `sigstore-kms-[name]`, like `sigstore-kms-my-hsm`. They can be invoked with cosign like `cosign [sub-command] --key "my-hsm://my-key-id" ...
```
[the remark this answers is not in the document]
ramonpetgrave64: For now, if there is a clash, it will run the one that appears first in PATH. It won't error due to a clash. We could allow the user to specify the location of the executable with an env variable.

test/cliplugin/localkms/main.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/handler"
+)
+
+const expectedProtocolVersion = "1"
```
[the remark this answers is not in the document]
ramonpetgrave64: You mean for the main program to invoke the plugin program with a list of protocol versions?

No, we only want the sigstore library to support one version of the protocol at a time (latest). Plugin authors may want to implement their own backwards compatibility though branching and using Dispatch with different versions of their `SignerVerifier` implementations.

pkg/signature/kms/cliplugin/signer.go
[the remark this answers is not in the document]
ramonpetgrave64: @loosebazooka No, it's the other way. `PluginClient.CreateKey()` via `invokePlugin()` and `Dispatch()` will always call `handler.CreateKey()`. Yes, it will inject timeouts. Authors of other languages would only have to look at `handler` as a reference.

@haydentherapper We could put all the content of `handler.CreateKey()` into `handler.Dispatch()`, but then it seems like a lot of behavior for one function.

test/cliplugin/localkms/implementation.go
ramonpetgrave64: added.

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: fixed.

test/cliplugin/localkms/implementation.go
ramonpetgrave64: added.

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: fixed.

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: fixed.

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,83 @@
...
+* arg 2
+  * JSON of initialization options and method arguments.
+* stdout
+  * JSON of method return values.
+
+See [./common/interface.go](./common/interface.go) for the full JSON schema.
+
+### Implementation
+
+Plugin authors must implement the `kms.SignerVerifier` interface methods in their chosen language. Each method will inbvoke your program once, and the reponse will be parsed from stdout.
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+Adding new methods or options are *not* necessarily breaking changes to the schemas, so we may consider these to be minor version increments, both to the protocol version and the sigstore version.
+
+### Removing Methods
+
+Removing methods, or altering their signatures will break the schemas and will require major version increments, both to the protocl version and the sigstore version.
+
+### Example Plugin
+
+We have an example plugin in [test/cliplugin/localkms](../../../.././test/cliplugin/localkms).
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed.

pkg/signature/kms/cliplugin/client.go
```diff
@@ -0,0 +1,59 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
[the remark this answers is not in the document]
ramonpetgrave64: thanks, fixed.

test/cliplugin/localkms/README.md
ramonpetgrave64: fixed.

pkg/signature/kms/cliplugin/signer_program_test.go
ramonpetgrave64: fixed.

test/cliplugin/localkms/README.md
ramonpetgrave64: fixed.

ramonpetgrave64:
@goatwu1993 We're planning to implement SignMessage within the next 1 or 2 small PRs. First, we want to agree on a good structure.

> Our company's use case.
> 
> client: forked cosign

I'm happy you're interested in having your own custom plugin! Since you're using cosign as part of a web server, can you share some more about your use case and performance requirements?

> the full payload can be sent to the plugin?

Yes, we intend to send the full file content/message/payload to the plugin over stdin.

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: A new method (with new version) might break the plugin programs, if the plugin does not respect the new protocol version or the new method is not implemented by the plugin.

It's plausible that a new method might be rarely used, like `CreateKey()`, for example.

Added a TODO to implement semver, probably with https://pkg.go.dev/golang.org/x/mod/semver

pkg/signature/kms/cliplugin/client.go
```diff
@@ -0,0 +1,59 @@
...
+// init registers the plugin system as a provider. It does not search for plugin programs.
+// users must import this package, e.g.,  `import _ "github.com/sigstore/sigstore/pkg/signature/kms/cliplugin"`
+func init() {
+	kms.AddProvider(kms.CLIPluginProviderKey, func(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts ...signature.RPCOption) (kms.SignerVerifier, error) {
+		return LoadSignerVerifier(ctx, keyResourceID, hashFunc, opts...)
+	})
+}
+
+// LoadSignerVerifier creates a PluginClient with these InitOptions.
+func LoadSignerVerifier(ctx context.Context, inputKeyresourceID string, hashFunc crypto.Hash, opts ...signature.RPCOption) (kms.SignerVerifier, error) {
```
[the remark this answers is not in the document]
ramonpetgrave64: Because it has a vault token? 
* https://github.com/ramonpetgrave64/sigstore/blob/185deaa2f5438a6e8ba828f625fee29724e3895c/pkg/signature/options/rpcauth.go#L28

I agree about not putting secrets in CLI, But I think we should keep `RPCOption`, because it has the `Context` and `KeyVersion`, and probably drop `options.RPCAuth`.
* https://github.com/ramonpetgrave64/sigstore/blob/185deaa2f5438a6e8ba828f625fee29724e3895c/pkg/signature/options.go#L31

And the hashivault package, is already written to support reading creds from env, same as the other KMS packages.

pkg/signature/kms/cliplugin/common/interface.go
ramonpetgrave64: We could, but it still requires invoking the plugin first. And it seems like `DefaultAlgorithm()` is used by cosign only when calling `CreateKey()`.

pkg/signature/kms/cliplugin/signer.go
ramonpetgrave64: see https://github.com/sigstore/sigstore/pull/1901#discussion_r1893749368

test/cliplugin/localkms/main.go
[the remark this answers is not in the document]
ramonpetgrave64: Added a dispatch/README.md

goatwu1993:
No i use cosign to send the payload to the signing server

example. 
https://github.com/goatwu1993/example-cosign-server/tree/main

test/cliplugin/localkms/main.go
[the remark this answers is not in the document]
loosebazooka: I thought this was the example plugin implementaiton? Anyway, I think I got my answer.

pkg/signature/kms/cliplugin/common/interface.go
haydentherapper: You're right, I was thinking this was something that could be statically defined across plugins but each plugin has their own default algorithm.

I think there are some improvements we can make outside of this PR to change the behavior of `CreateKey` to use a default if no algorithm is specified, but given the other thread about deprecating it entirely, we can remove `DefaultAlgorithm` at the same time.

pkg/signature/kms/cliplugin/common/interface.go
haydentherapper: I might be misunderstanding this. I should this would be for the timeout for each call from the client to the plugin. If it's long-running, that should be fine as each timeout is computed for that invocation.

In the spirit of simplifying what plugin authors need to create, I think keeping these methods small in scope would be good. If we later need to add per-method timeouts, that's a non-breaking change.

haydentherapper (commented):
LGTM, just need to fix a test!

pkg/signature/kms/cliplugin/common/interface.go
ramonpetgrave64: The CtxDeadline is not a duration, it’s a specific timestamp in the future. But I think we should fully implement the interface, and plugin authors can choose to ignore the CtxDeadline. Our CommandContext invocation will respect a CtxDeadline, if the caller provides one.

ramonpetgrave64: But since the plugin Programs themselves are not long-running, and author can interpret InitOptions.CyxDeadline, as a deadline for doing cold-start initialization tasks, like fetching a TUF cache, for example. 

I’ll add a TODO to add some guidance about handling Deadlines and InitOptions.

pkg/signature/kms/cliplugin/common/interface.go
ramonpetgrave64: We could consider removing the method from the plugin as a breaking change. There may be people that use it, but we don’t have the visibility on their use cases.

ramonpetgrave64:
> LGTM, just need to fix a test!

Thanks, fixed the test.

pkg/signature/kms/cliplugin/common/interface.go
haydentherapper: I think we should document what is required (namely, SignMessage), and what is optional, so plugin authors know where to focus their development.

We can circle back to this before we cut a release with this plugin, and see about removing functionality from Cosign.

pkg/signature/kms/cliplugin/common/interface.go
haydentherapper: Sounds good - related to the other comment, let's document what is a must for handling and what is optional.

haydentherapper (approved):
Will leave open in case anyone else has comments, can merge next week.

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,91 @@
...
+  * Data to be signed or verified.
+* arg 1
+  * A number identifying the version of this protocol.
+    * In the future we may change this protocol, either the encoding or the formatting of argument and return values.
+    Protocol changes will result in a major-version bump for the library.
+    When changing the protocol, new versions of sigstore will not maintain backwards compatibility with
+    previous protocol versions. If a plugin author wishes, they may branch their plugin program’s behaviour
+    to be compatible with multiple versions of the protocol, or multiple major versions of sigstore and cosign.
+* arg 2
+  * JSON of initialization options and method arguments.
```
bobcallaway, line 29: why could these not be combined?

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,91 @@
...
+* stdout
+  * JSON of method return values.
+
+See [./common/interface.go](./common/interface.go) for the full JSON schema.
+
+The plugin program must first exit before sigstore begins parsing responses.
```
bobcallaway, line 35: why is that? In general I dont think you could enforce that

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Error Handling
+
+The plugin program’s stderr will be redirected to the main program’s stderr. This way, the main program may also see the plugin program’s debug messages.
+
+Plugin authors may return errors with `PluginResp.ErrorMessage`, but the plugin's exit status will be ignored.
+
+### Implementation
+
+Plugin authors must implement the `kms.SignerVerifier` interface methods in their chosen language. Each method will invoke your program once, and the reponse will be parsed from stdout.
```
bobcallaway, line 45:
```suggestion
Plugin authors must implement the `kms.SignerVerifier` interface methods in their chosen language. Each method will invoke your program once, and the response will be parsed from stdout.
```

also it needs to logically implement the interface, not strictly (since you're not passing anything Go-specific over the wire from cosign/other tools?

pkg/signature/kms/cliplugin/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+Changes to the `SignerVerifier` interface are to be handled in [./signer.go's](./signer.go) `PluginClient` and [./handler/dispatch.go's](./handler/dispatch.go) `Dispatch()`.
+
+### Adding New Methods or Method Options
+
+Adding new methods or options are *not* necessarily breaking changes to the schemas, so we may consider these to be minor version increments, both to the protocol version and the sigstore version.
+
+### Removing Methods
+
+Removing methods, or altering their signatures will break the schemas and will require major version increments, both to the protocl version and the sigstore version.
```
bobcallaway, line 61:
```suggestion
Removing methods, or altering their signatures will break the schemas and will require major version increments, both to the protocol version and the sigstore version.
```

pkg/signature/kms/cliplugin/client.go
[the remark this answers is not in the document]
bobcallaway: I think that's something we could explore, but would require us to write our own AWS/MSFT/GOOG/HV binaries.

pkg/signature/kms/cliplugin/handler/dispatch.go
```diff
@@ -0,0 +1,84 @@
...
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/common"
+)
+
+var (
+	ErrorPluginArgsParse   = errors.New("error parsing plign args")
```
bobcallaway, line 30:
```suggestion
	ErrorPluginArgsParse   = errors.New("error parsing plugin args")
```

pkg/signature/kms/cliplugin/handler/dispatch.go
```diff
@@ -0,0 +1,84 @@
...
+	return &args, nil
+}
+
+// WriteResponse writes JSON-serialized PluginResp to the output.
+func WriteResponse(stdout io.Writer, resp *common.PluginResp) error {
+	enc, err := json.Marshal(resp)
+	if err != nil {
+		return err
+	}
+	fmt.Fprint(stdout, string(enc))
```
bobcallaway, line 53: if the first arg would ever be something other than `os.Stdout` I'd recommend renaming that writer as `w` or something else. If it will always be `os.Stdout`, why pass it as an arg?

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: Currently, we invoke the plugin program with `Cmd.Output()`, which waits until the plugin program exits, because:
1. We need to make sure that the main program received all of the intended output from the plugin program.
2. For the sake of users we may want to guarantee that plugin programs do eventually exit.

If we don't wait for the plugin program to exit, the protocol will need to be slightly more complex: read a full line of stdout before parsing the `PluginResp`. The newline character serves as a delimeter.

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: It's possible we use a very different protocol than the current version. In order to keep things simple for future plugin programs, I think we should keep this first argument simple to parse.

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: fixed

pkg/signature/kms/cliplugin/handler/dispatch.go
ramonpetgrave64: fixed

pkg/signature/kms/cliplugin/README.md
ramonpetgrave64: fixed

pkg/signature/kms/cliplugin/handler/dispatch.go
haydentherapper: Can we change the var name to `w`?

pkg/signature/kms/cliplugin/handler/dispatch.go
```diff
@@ -0,0 +1,84 @@
...
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/common"
+)
+
+var (
+	ErrorPluginArgsParse   = errors.New("error parsing pluign args")
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

haydentherapper (approved):

test/e2e/vault.Dockerfile
```diff
@@ -13,4 +13,4 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-FROM docker.io/hashicorp/vault:1.18.2@sha256:2090eb7ac7a4bdef802f685698bd4dc0740de683affe8ff7df55f4fc77077ba7
+FROM docker.io/hashicorp/vault:1.18.3@sha256:8f1ba670da547c6af67be55609bd285c3ee3d8b73f88021adbfc43c82ca409e8
```
haydentherapper, line 16: looks like a few unrelated changes got picked up? here and the fuzz go.mod/sum changes

pkg/signature/kms/kms.go
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2021 The Sigstore Authors.
+// Copyright 2024 The Sigstore Authors.
```
haydentherapper, line 2: copyright should not be updated for an existing file.

test/e2e/vault.Dockerfile
ramonpetgrave64: fixed

pkg/signature/kms/kms.go
ramonpetgrave64: fixed

haydentherapper (dismissed):

haydentherapper (approved):
This looks great!

haydentherapper (approved):

PR #1918, opened 2024-12-26
cliplugin: add mocks and serialization testing

ramonpetgrave64:
Followup to https://github.com/sigstore/sigstore/pull/1901

#### Summary

Adds testing each of the `PluginClient` methods and the functions in the `handler` package by mocking and simulating the behavior of `invokePlugin`.

#### Release Note

* Add mocks and serialization tests for `PluginClient`.

#### Documentation

Added some more details on developing and testing plugins.


ramonpetgrave64:
@haydentherapper @loosebazooka

pkg/signature/kms/cliplugin/client.go
```diff
@@ -60,7 +60,7 @@ func LoadSignerVerifier(ctx context.Context, inputKeyResourceID string, hashFunc
 	if deadline, ok := ctx.Deadline(); ok {
 		initOptions.CtxDeadline = &deadline
 	}
-	pluginClient := newPluginClient(executable, initOptions)
+	pluginClient := newPluginClient(executable, initOptions, makeCmd)
```
loosebazooka, line 63: Maybe this is my confusion with go paradigms, but is there away to avoid this? Seems like we're only exposing this for testing? Do we expect some plugin client to actually provide a non-default version of this?

loosebazooka (approved):
looks fine to me

pkg/signature/kms/cliplugin/signer_test.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	if diff := cmp.Diff(testContextDeadline, ctxDeadline); diff != "" {
+		s.t.Errorf("unexpected context deadline (-want +got):\n%s", diff)
+	}
+	return testPublicKey, nil
+}
+
+// TestPluginClient tests each of PluginClient's methods for correct encoding and decoding between a simulated plugin program,
+// by mocking the Cmd function and using TestSignerVerifierImpl to both check and return expected values.
+func TestPluginClient(t *testing.T) {
```
haydentherapper, line 254: What is the test for? It looks like all calls are mocked so it's mainly testing that an interface is implemented. Am I missing something?

pkg/signature/kms/cliplugin/client.go
haydentherapper: I was looking around blogs and stackoverflow to see if there were any other suggestions. It seems like the only other Go-ish approach is a global variable, though I don't think that's much cleaner.

haydentherapper (dismissed):
Does https://github.com/sigstore/sigstore/blob/2019c85f48e18f9c821a04972ca7d9e456340092/pkg/signature/kms/cliplugin/signer_program_test.go#L84-L85 need to be updated? I think there is still value in this as an e2e test, so can the TODO be removed?

pkg/signature/kms/cliplugin/signer_test.go
ramonpetgrave64: It's to check that, for each method, all the arguments and return values can survive the JSON serialization and deserialization that happens through invokePlugin.

pkg/signature/kms/cliplugin/signer_test.go
```diff
@@ -0,0 +1,306 @@
+//
+// Copyright 2024 The Sigstore Authors.
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

ramonpetgrave64:
> ... need to be updated? I think there is still value in this as an e2e test, so can the TODO be removed?

Yes, but I just moved the deadline test cases into a separate `TestInvokePluginPreBuilt()`.

haydentherapper (approved):

PR #1919, opened 2024-12-30
kms plugin: add SignMessage

ramonpetgrave64:
pending to https://github.com/sigstore/sigstore/pull/1918

Compare the branch diffs: https://github.com/ramonpetgrave64/sigstore/compare/ramonpetgrave64-kms-plugin-cli-mocks...ramonpetgrave64:sigstore:ramonpetgrave64-kms-plugin-cli-signmessage

#### Summary

Adds the `SignMessage()` method to the PluginClient.
Adds a test against the json serialization of PluginArgs and PluginResp.

#### Documentation

Added link to example JSON renderings of PluginArgs and PluginResp.

## TODO

- [x] update the examples in README to use SignMessage, instead of CereateKey.


pkg/signature/kms/cliplugin/common/interface.go
```diff
@@ -85,3 +88,13 @@ type CreateKeyArgs struct {
 type CreateKeyResp struct {
 	PublicKeyPEM []byte `json:"publicKeyPEM"`
 }
+
+// SignMessageArgs contains the serialized arguments for `SignMessage()`.
+type SignMessageArgs struct {
```
haydentherapper, line 93: Does this need a `CtxDeadline`? Same with `DefaultAlgorithmArgs`

pkg/signature/kms/cliplugin/handler/options.go
```diff
@@ -0,0 +1,68 @@
...
+	if commonOpts.KeyVersion != nil {
+		opts = append(opts, options.WithKeyVersion(*commonOpts.KeyVersion))
+	}
+	if commonOpts.RemoteVerification != nil {
+		opts = append(opts, options.WithRemoteVerification(*commonOpts.RemoteVerification))
+	}
+	return opts
+}
+
+// getRPCOptions builds the []signature.MessageOption from common.MessageOptions.
```
haydentherapper, line 46:
```suggestion
// unpackMessageOptions builds the []signature.MessageOption from common.MessageOptions.
```

pkg/signature/kms/cliplugin/handler/options_test.go
```diff
@@ -0,0 +1,155 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package handler implements helper functions for plugins written in go.
+// It parses arguments  and return values to and from the supplied `SignerVerifier` implementation.
+
+//go:build !signer_program
```
haydentherapper, line 19: why does this have to be excluded?

pkg/signature/kms/cliplugin/options.go
```diff
@@ -0,0 +1,82 @@
...
+	"crypto"
+	"time"
+
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/common"
+)
+
+// packRPCOptions extracts properties of all of opts into struct ready for serializing.
+func packRPCOptions(opts []signature.RPCOption) *common.RPCOptions {
+	ctx := context.TODO()
```
haydentherapper, line 30: We should be opinionated on a default context, setting some reasonable timeouts. I think `context.Background()` should also be used rather than `TODO`.

pkg/signature/kms/cliplugin/options.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+	"crypto"
+	"time"
+
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin/common"
+)
+
+// packRPCOptions extracts properties of all of opts into struct ready for serializing.
+func packRPCOptions(opts []signature.RPCOption) *common.RPCOptions {
```
haydentherapper, line 29: Could these functions be in the same file as the unpack.*Options?

haydentherapper (commented):
looks good, just a few small comments

pkg/signature/kms/cliplugin/handler/options_test.go
ramonpetgrave64: That's before I found a way to not include them with the `make test-signer-program` runs. I changed `./...` to `./`

pkg/signature/kms/cliplugin/handler/options.go
ramonpetgrave64: fixed

pkg/signature/kms/cliplugin/options.go
ramonpetgrave64: we could, but it could lead to a circular import situation in the future.

pkg/signature/kms/cliplugin/options.go
ramonpetgrave64: I agree, but the opinion should come from our primary caller cosign, which does set a 3minute timeout. 
Changed the contexts to Background.

Makefile
```diff
@@ -58,7 +58,7 @@ test: ## Run Tests for all Go modules.
 test-signer-program: ## Run Tests for the cliplugin against a pre-compiled plugin program.
 	set -o xtrace; \
 		go -C $(TEST_LOCALKMS_DIR) build -o $(TEST_LOCALKMS_BIN_DIR)/sigstore-kms-testkms && \
-		go -C $(CLI_PLUGIN_DIR) test -tags=signer_program ./... -key-resource-id testkms://$(TEST_LOCALKMS_BIN_DIR)/key.pem
+		go -C $(CLI_PLUGIN_DIR) test -tags=signer_program ./ -key-resource-id testkms://$(TEST_LOCALKMS_BIN_DIR)/key.pem
```
ramonpetgrave64, line 61: make sure we only run tests at the package root, not in sub packages

pkg/signature/kms/cliplugin/common/interface.go
ramonpetgrave64: No, because in `SignerVerifier`, this method's context is in its opts argument.

pkg/signature/kms/cliplugin/options.go
haydentherapper: In what way? This effectively is marshal and unmarshal, so grouping them together seems logically structured.

ramonpetgrave64: When I was trying to think of a way for the kms package to import the cliplugin package directly, I thought the handler package might need to depend on the clipluign package, and maybe causing a circular dependency. But that won't be the case, so this isn't an issue.

I moved the pack and unpack functions to a new encoding package.

pkg/signature/kms/cliplugin/common/options.go
```diff
@@ -31,13 +36,16 @@ type SignOptions struct {
...
 // The plugin program should instead read secrets with env variables.
 type RPCOptions struct {
+	// CtxDeadline serializes to RFC 3339. See https://pkg.go.dev/time@go1.23.5#Time.MarshalJSON. e.g, 2025-04-01T02:47:00Z.
 	CtxDeadline        *time.Time `json:"ctxDeadline,omitempty"`
 	KeyVersion         *string    `json:"keyVersion,omitempty"`
 	RemoteVerification *bool      `json:"remoteVerification,omitempty"`
 }
 
 // MessageOptions contains the values for signature.MessageOption.
 type MessageOptions struct {
```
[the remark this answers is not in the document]
ramonpetgrave64: Yes, the artifact will be sent over stdin to the plugin program

pkg/signature/kms/cliplugin/encoding/options.go
```diff
@@ -0,0 +1,126 @@
...
+		RemoteVerification: &remoteVerification,
+	}
+}
+
+// UnpackRPCOptions builds the []signature.RPCOption from common.RPCOptions.
+func UnpackRPCOptions(commonOpts *common.RPCOptions) []signature.RPCOption {
+	opts := []signature.RPCOption{}
+	if commonOpts.CtxDeadline != nil {
+		ctx, cancel := context.WithDeadline(context.Background(), *commonOpts.CtxDeadline)
+		defer cancel()
```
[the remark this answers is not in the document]
ramonpetgrave64: You're right. We shouldn't defer here. But since this code will be called by the plugin program, I think this means we should discard the cancel func created here, since there's no reason for the dispatch package to want to cancel the context early. If a user wants to check if the deadline was exceeded, then they can call `ctx.Err()`.

wdyt?

pkg/signature/kms/cliplugin/common/interface.go
```diff
@@ -77,11 +82,24 @@ type DefaultAlgorithmResp struct {
...
+	SignOptions *SignOptions `json:"signOptions"`
+}
+
+// SignMessageResp contains the serialized response for `SignMessage()`.
+type SignMessageResp struct {
+	// Signature is a base64 encoding of the Public Key PEM bytes. e.g, []byte("any-signature") serializes to "W55LXNpZ25hdHVyZQ==".
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

pkg/signature/kms/cliplugin/encoding/options.go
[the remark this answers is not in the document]
haydentherapper: I agree, discard sounds good.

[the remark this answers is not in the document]
ramonpetgrave64: fixed.

haydentherapper (approved):

PR #1933, opened 2025-01-14
use ubuntu-22.04 for the e2e tests workflow 

cpanato:
- use ubuntu-22.04 for the e2e tests workflow 
- drop obsolete version

we are getting those errors
```
panic: [launcher] Failed to get the debug url: [9175:9175:0114/132426.963765:FATAL:zygote_host_impl_linux.cc(126)] No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/main/docs/linux/suid_sandbox_development.md for more information on developing with the SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.
```

see https://github.com/sigstore/sigstore/actions/runs/12753677837/job/35573189089?pr=1932 as example

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -27,7 +27,10 @@ permissions:
 jobs:
   e2e_test:
     name: e2e integration tests
-    runs-on: ubuntu-latest
+    # using ubuntu-22.04 because on 24.02 we are getting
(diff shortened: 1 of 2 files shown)


haydentherapper (approved):

bobcallaway:
i think we might be able to update the apparmor rules here and stay on latest

cpanato:
> i think we might be able to update the apparmor rules here and stay on latest

Is it okay if we investigate that in a follow-up?

bobcallaway:
> > i think we might be able to update the apparmor rules here and stay on latest
> 
> Is it okay if we investigate that in a follow-up?

done: https://github.com/sigstore/sigstore/pull/1934

PR #1946, opened 2025-01-27
cliplugin: add SupportedAlgorithms(), PublicKey(), and CryptoSigner().

ramonpetgrave64:
#### Summary

pending #1944, compare branch diffs:

 - https://github.com/ramonpetgrave64/sigstore/compare/ramonpetgrave64-kms-plugin-cli-verifysignature...ramonpetgrave64:sigstore:ramonpetgrave64-kms-plugin-cli-cryptosigner-working

Adds the methods for SupportedAlgorithms(), PublicKey(), and CryptoSigner().

The next PRs will address remaining TODOs.

#### Release Note

add PluginClient methods SupportedAlgorithms(), PublicKey(), and CryptoSigner().

#### Documentation

Add note about the CryptoSigner method to the readme.

PR #1947, opened 2025-01-28
cliplugin: use caller contexts

ramonpetgrave64:
#### Summary

pending #1946. See branch diff: https://github.com/ramonpetgrave64/sigstore/compare/ramonpetgrave64-kms-plugin-cli-cryptosigner-working...ramonpetgrave64:sigstore:ramonpetgrave64-kms-plugin-cli-context

Makes PluginClient methods use the contexts of the original caller when calling invokePlugin. The benefit is that the Cmd can be cancelled using the caller's cancel func.

#### Release Note

PluginClient methods can be cancelled with the caller's context cancel func.


pkg/signature/kms/cliplugin/signer_test.go
```diff
@@ -187,26 +187,36 @@ func TestInvokePlugin(t *testing.T) {
...
-				} else if diff := cmp.Diff(testPluginArgs, pluginArgs); diff != "" {
-					t.Errorf("parsing plugin args (-want +got):\n%s", diff)
+				outputFunc := func() ([]byte, error) {
+					if err := ctx.Err(); err != nil {
+						return nil, err
+					}
+
+					if err := ctx.Err(); err != nil {
+						t.Error(err)
+					}
```
bobcallaway, line 197: is there a reason this is duplicated?

ramonpetgrave64: no, fixed.

pkg/signature/kms/cliplugin/signer.go
```diff
@@ -213,6 +207,9 @@ type CryptoSigner struct {
 
 // CryptoSigner returns a wrapper around PluginClient.
 func (c PluginClient) CryptoSigner(ctx context.Context, errFunc func(error)) (crypto.Signer, crypto.SignerOpts, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, nil, err
+	}
```
bobcallaway, line 212: it feels off to check this at the time of object creation instead of object usage.

pkg/signature/kms/cliplugin/signer.go
```diff
@@ -144,10 +144,8 @@ func (c PluginClient) PublicKey(opts ...signature.PublicKeyOption) (crypto.Publi
 		},
 	}
 	ctx := context.Background()
-	if deadline := args.PublicKey.PublicKeyOptions.RPCOptions.CtxDeadline; deadline != nil {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithDeadline(ctx, *deadline)
-		defer cancel()
+	for _, opt := range opts {
+		opt.ApplyContext(&ctx)
```
haydentherapper, line 148: To confirm, is the expectation that the creator of the context will call `defer cancel()`?

ramonpetgrave64: Yes, or they may call it `cancel()` early.

pkg/signature/kms/cliplugin/signer.go
ramonpetgrave64: I'm thinking of this context as having a deadline that's specific for object initialization. If the caller wants to reuse the same context in method calls, then that's okay.

- https://github.com/sigstore/timestamp-authority/blob/3c5d80434f3235c52aca136de038406b24a95b83/cmd/fetch-tsa-certs/fetch_tsa_certs.go#L88

haydentherapper: For what reason would this have an error at this point?

haydentherapper:
Just a meta comment: The commit history seems to be pulling in tons of old commits. It's not a big deal since we'll squash the commit before merge, but it's hard to follow the changes in this PR, so I'd suggest rebasing off main as we merge PRs.

pkg/signature/kms/cliplugin/signer.go
ramonpetgrave64: If the context already expired, then no reason to continue trying to initialize the CrytpoSigner.

Also this article talks about using contexts in initializer functions, and that reusing it for methods leads to confusion.

 - https://go.dev/blog/context-and-structs

ramonpetgrave64: Also, does it make sense to call the `errFunc` here?

haydentherapper (approved):

PR #1948, opened 2025-01-28
cliplugin: semver, add tests for hash func encoding

ramonpetgrave64:
#### Summary

pending #1947, branch diff: https://github.com/ramonpetgrave64/sigstore/compare/ramonpetgrave64-kms-plugin-cli-context...ramonpetgrave64:sigstore:ramonpetgrave64-kms-plugin-cli-hashencoding

Adds tests against know hash func and their JSON encoding.
The current set of hash funcs and their encoding has not changed in ~8 years, but they might change in the future, since they are ordered by iota.

 - https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/crypto/crypto.go;l=68-89;bpv=0

This PR 

 - adds a test that can give us a warning if and when they do change, and we would likely need a major version bump of the protocol version.
 - adds semver to the plugin protocol version at "v1.0.0".

#### Release Note

The plugin protocol version now uses semver at "v1.0.0".

#### Documentation

No new documentation changes


pkg/signature/kms/cliplugin/common/interface.go
```diff
@@ -26,8 +26,7 @@ const (
 	// Breaking changes to the PluginClient and this schema necessarily mean major version bumps of
 	// this ProtocolVersion and the sigstore version.
 	// Plugin authors may choose to be backwards compatible with older versions.
-	// TODO: change this to be semver compatible, like v1, or v1.0.0.
-	ProtocolVersion               = "1"
+	ProtocolVersion               = "v1.0.0"
```
haydentherapper, line 29: I think this amount of semver specificity is overkill and a single major version should be sufficient. Clients don't need to be aware of non-breaking changes to the protocol.

ramonpetgrave64: changed

PR #1956, opened 2025-02-05
cliplugin: convert module to package only

ramonpetgrave64:
#### Summary

Pending #1948, #1958

branch diff: 
 - https://github.com/ramonpetgrave64/sigstore/compare/ramonpetgrave64-kms-plugin-cli-hashencoding...ramonpetgrave64:sigstore:no-module
 - https://github.com/ramonpetgrave64/sigstore/compare/lint-fix...ramonpetgrave64:sigstore:no-module

Moves the details of `kms.signerVerifier` to a new bridge package `signerverifier.SignerVerifier` to avoid import cycles.

Marks `ProviderNotFoundError` for deprecation.

Converts cliplugin to be a package only, not a module. This lets consumers, such as cosign, avoid having to load the module like below, and its okay because there are no new dependencies.

```
import _ "github.com/sigstore/sigstore/pkg/signature/kms/cliplugin"
```

To start using cliplugin, consumers need only bump the version of sigstore.

- [x] lint fixes, becuase golanci-lint does not run against sub-modules.

### Testing

Example error

```
error during command execution: signing ../blob.txt: reading key: kms get: exec: "sigstore-kms-myawskmsX": executable file not found in $PATH
```

#### Release Note

cliplugin is now a package only, not a module that needs to be loaded for initialization. There is no need to explicitly import cliplugin.

#### Documentation

No doc changes.


ramonpetgrave64:
Lint errors should be fixed after merging #1958. Afterwards, we can remove that PR's explicit commands to run golangci-lint the the cliplugin directory.

ramonpetgrave64:
@haydentherapper @bobcallaway

haydentherapper:
note: the stuck CI is expected, we'll remove that as a requirement

ramonpetgrave64:
@haydentherapper It may be because the job's changed names, because of the matrix?

ramonpetgrave64:
I re-did the commits to follow a better flow.

pkg/signature/kms/kms.go
```diff
@@ -69,14 +67,15 @@ func Get(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts .
...
-	return nil, &ProviderNotFoundError{ref: keyResourceID}
+	sv, err := cliplugin.LoadSignerVerifier(ctx, keyResourceID, hashFunc, opts...)
+	// We don't return a ProviderNotFoundError because cosign currently interprets those in a confusing way:
+	// "error during command execution: signing ../blob.txt: reading key: loading URL: unrecognized scheme: mykms://".
+	// Instead, without the ProviderNotFoundError, we would get:
+	// "error during command execution: signing ../blob.txt: reading key: kms get: exec: "sigstore-kms-mykms": executable file not found in $PATH"
+	// if errors.Is(err, exec.ErrNotFound) {
+	// 	return nil, &ProviderNotFoundError{ref: keyResourceID}
+	// }
+	return sv, err
```
bobcallaway, line 78:
```suggestion
	// We don't return a ProviderNotFoundError because cosign currently interprets those in a confusing way:
	// "error during command execution: signing ../blob.txt: reading key: loading URL: unrecognized scheme: mykms://".
	// Instead, without the ProviderNotFoundError, we would get:
	// "error during command execution: signing ../blob.txt: reading key: kms get: exec: "sigstore-kms-mykms": executable file not found in $PATH"
	// if errors.Is(err, exec.ErrNotFound) {
	// 	return nil, &ProviderNotFoundError{ref: keyResourceID}
	// }
	return cliplugin.LoadSignerVerifier(ctx, keyResourceID, hashFunc, opts...)
```

pkg/signature/kms/kms_test.go
```diff
@@ -0,0 +1,42 @@
...
+	testHashFunc := crypto.SHA256
+	testCtx := context.Background()
+	testKey := "gundam://00"
+
+	_, err := Get(testCtx, testKey, testHashFunc)
+
+	// exec.ErrNotFound is returned by cliplugin.LoadSignerVerifier().
+	if !errors.Is(err, exec.ErrNotFound) {
+		t.Errorf("wanted exec.ErrNotFound, got: %v", err)
+	}
```
bobcallaway, line 41:
```suggestion
	// exec.ErrNotFound is returned by cliplugin.LoadSignerVerifier().
	if _, err := Get(testCtx, testKey, testHashFunc); !errors.Is(err, exec.ErrNotFound) {
		t.Errorf("wanted exec.ErrNotFound, got: %v", err)
	}
```

pkg/signature/kms/signerverifier/interface.go
```diff
@@ -0,0 +1,33 @@
+//
+// Copyright 2022 The Sigstore Authors.
```
bobcallaway, line 2:
```suggestion
// Copyright 2025 The Sigstore Authors.
```

test/cliplugin/localkms/go.mod
```diff
@@ -4,19 +4,16 @@ go 1.22.0
 
 replace github.com/sigstore/sigstore => ../../..
 
-replace github.com/sigstore/sigstore/pkg/signature/kms/cliplugin => ../../../pkg/signature/kms/cliplugin
-
 require (
 	github.com/davecgh/go-spew v1.1.1
-	github.com/sigstore/sigstore v1.8.10
-	github.com/sigstore/sigstore/pkg/signature/kms/cliplugin v0.0.0-00010101000000-000000000000
+	github.com/sigstore/sigstore v0.0.0-00010101000000-000000000000
```
bobcallaway, line 9: this looks strange, perhaps it doesn't matter with the `replace` on line 5?

ramonpetgrave64: That's right. Pinning an actual version won't be useful, I think.

pkg/signature/kms/kms.go
ramonpetgrave64: changed

pkg/signature/kms/kms.go
```diff
@@ -23,15 +23,14 @@ import (
...
-	// CLIPluginProviderKey is a placeholder used to register the cliplugin as a provider in AddProvider().
-	// Its value should not conflict with any potential callers of AddProvider().
-	CLIPluginProviderKey = "cliplugin"
+	"github.com/sigstore/sigstore/pkg/signature/kms/cliplugin"
+	"github.com/sigstore/sigstore/pkg/signature/kms/signerverifier"
 )
 
 // ProviderNotFoundError indicates that no matching KMS provider was found
+//
+// Deprecated: ProviderNotFoundError is no longer rerturned by Get(). If no matching provider is found
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

pkg/signature/kms/signerverifier/interface.go
ramonpetgrave64: fixed

pkg/signature/kms/kms.go
```diff
@@ -69,14 +67,15 @@ func Get(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts .
...
 	}
-	if pi, ok := providersMap[CLIPluginProviderKey]; ok {
-		sv, err := pi(ctx, keyResourceID, hashFunc, opts...)
-		if err != nil {
-			return nil, err
-		}
-		return sv, nil
-	}
-	return nil, &ProviderNotFoundError{ref: keyResourceID}
+	// We don't return a ProviderNotFoundError because cosign currently interprets those in a confusing way:
```
haydentherapper, line 70: This library is used outside of Cosign so we shouldn't be making changes just because of how Cosign outputs this error.

Isn't it more accurate to log that it is an unrecognized scheme rather than a missing executable? Whether the KMS provider is an executable or module is a detail the user doesn't need to be aware of.

I think we can get the best of both worlds - wrap the original error from `cliplugin.LoadSignerVerifier` in a `ProviderNotFoundError`.

pkg/signature/kms/kms.go
```diff
@@ -89,10 +88,7 @@ func SupportedProviders() []string {
 }
 
 // SignerVerifier creates and verifies digital signatures over a message using a KMS service
+// It is a wrapper for signerverifier.SignerVerifier.
 type SignerVerifier interface {
```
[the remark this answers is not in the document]
ramonpetgrave64: That works! though we have to copy the contents of the interface. It's okay, because if the two interfaces are ever out of sync, I see we would get a compiler error since `cliplugin.LoadsignerVerifier()'`s return value would not satisfy the return value of `Get()`.

pkg/signature/kms/kms.go
ramonpetgrave64: changed, and added tests

haydentherapper (approved):

PR #1980, opened 2025-02-14
pkg/signature/kms: revert Get to return ProviderNotFoundError

### LINKED ISSUE #1979: cliplugin changes break cosign tests
**Description**
I was trying to update cosign sigstore/sigstore dependency and I ran into a few test failures that seems to be related to recent changes in https://github.com/sigstore/sigstore/pull/1956 . Those changes do not recognize anymore a "raw" keyref such as `/path/to/key` and thus a lot of cosign tests fail.

```
// getPluginExecutableAndKeyResourceID parses the inputKeyResourceID into the plugin executable and the actual keyResourceID.
func getPluginExecutableAndKeyResourceID(inputKeyResourceID string) (string, string, error) {
	parts := strings.SplitN(inputKeyResourceID, "://", 2)
	if len(parts) != 2 {
		return "", "", fmt.Errorf("%w: expected format: [plugin name]://[key ref], got: %s", ErrorInputKeyResourceID, inputKeyResourceID)
	}
	pluginName, keyResourceID := parts[0], parts[1]
	executable := PluginBinaryPrefix + pluginName
	return executable, keyResourceID, nil
}
```

This function now errors out when you provide a regular path without any provider, but it does not return `kms.ProviderNotFoundError`, thus cosign (and maybe other clients) interpret that as an hard-failure.


I think one fix could be to return a `kms.ProviderNotFoundError` in that function (maybe we can add a `msg` field to the error, if we want to keep the error message like it is now. 

Thoughts?


**Version**

```
cosign version                                                                                                                                                                                                                         ✘ 130 
  ______   ______        _______. __    _______ .__   __.
 /      | /  __  \      /       ||  |  /  _____||  \ |  |
|  ,----'|  |  |  |    |   (----`|  | |  |  __  |   \|  |
|  |     |  |  |  |     \   \    |  | |  | |_ | |  . `  |
|  `----.|  `--'  | .----)   |   |  | |  |__| | |  |\   |
 \______| \______/  |_______/    |__|  \______| |__| \__|
cosign: A tool for Container Signing, Verification and Storage in an OCI registry.

GitVersion:    2.4.2
GitCommit:     b6df9c777c365ce063a7e65075f2b08a3c76de2f
GitTreeState:  "clean"
BuildDate:     2025-02-04T19:41:37Z
GoVersion:     go1.23.6
Compiler:      gc
Platform:      darwin/arm64

```
but with `github.com/sigstore/sigstore v1.8.14` in go.mod.

ret2libc:
#### Summary
Closes #1979

### DIFF
--- a/pkg/signature/kms/kms.go
+++ b/pkg/signature/kms/kms.go
@@ -21,7 +21,6 @@ import (
 	"crypto"
 	"errors"
 	"fmt"
-	"os/exec"
 	"strings"
 
 	"github.com/sigstore/sigstore/pkg/signature"
@@ -30,11 +29,11 @@ import (
 
 // ProviderNotFoundError indicates that no matching KMS provider was found
 type ProviderNotFoundError struct {
-	ref string
+	Ref string
 }
 
 func (e *ProviderNotFoundError) Error() string {
-	return fmt.Sprintf("no kms provider found for key reference: %s", e.ref)
+	return fmt.Sprintf("no kms provider found for key reference: %s", e.Ref)
 }
 
 // ProviderInit is a function that initializes provider-specific SignerVerifier.
@@ -66,8 +65,9 @@ func Get(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts .
 		}
 	}
 	sv, err := cliplugin.LoadSignerVerifier(ctx, keyResourceID, hashFunc, opts...)
-	if errors.Is(err, exec.ErrNotFound) {
-		return nil, fmt.Errorf("%w: %w", &ProviderNotFoundError{ref: keyResourceID}, err)
+	var pnfErr *cliplugin.ProviderNotFoundError
+	if errors.As(err, &pnfErr) {
+		return nil, fmt.Errorf("%w: %v", &ProviderNotFoundError{Ref: keyResourceID}, err)
 	}
 	return sv, err
 }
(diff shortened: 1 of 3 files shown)


bobcallaway (approved):
lgtm, @ramonpetgrave64 ptal

bobcallaway (approved):
lgtm, @ramonpetgrave64 ptal

haydentherapper (approved):
LGTM, I missed this in my review

bobcallaway (approved):
lgtm, @ramonpetgrave64 ptal

pkg/signature/kms/kms.go
```diff
@@ -66,8 +65,9 @@ func Get(ctx context.Context, keyResourceID string, hashFunc crypto.Hash, opts .
 		}
 	}
 	sv, err := cliplugin.LoadSignerVerifier(ctx, keyResourceID, hashFunc, opts...)
-	if errors.Is(err, exec.ErrNotFound) {
```
ramonpetgrave64, line 69: I just tried this with both sigstore and cosign tests. I think the better way is to change this line, and you can also edit the sigstore tests to cover this case.

```
	if errors.Is(err, exec.ErrNotFound) || errors.Is(err, cliplugin.ErrorInputKeyResourceID) {
		return nil, fmt.Errorf("%w: %w", &ProviderNotFoundError{ref: keyResourceID}, err)
	}
```

ramonpetgrave64 (changes requested):

PR #2014, opened 2025-03-04
Implement default signing algorithms based on the key type

ret2libc:
#### Summary
As we make the Sigstore ecosystem more crypto agile, we realized that right now it is using some uncommon algorithms, like ECDSA-P384 with the SHA256 hash algorithm. However, we want the crypto-agility to be controlled and limited to a set of algorithms (see `AlgorithmRegistry`). Moreover, we want the agility to be mainly about supporting multiple key types, rather than multiple algorithms for the same key type. Thus, this PR introduces a single point where the defaults are defined.

This is going to simplify other changes across the Sigstore ecosystem, given that often enough the only information you have is the public key. Being able to derive the signing algorithm to use (or used to sign a blob) means we don't need to store or pass around any extra information (e.g. the hash function), because that can be derived automatically from the key.

See https://github.com/sigstore/sig-clients/issues/16 .

#### Release Note
- Introduce default `AlgorithmDetails` for public keys based on their types.

### DIFF
--- a/pkg/signature/algorithm_registry.go
+++ b/pkg/signature/algorithm_registry.go
@@ -21,6 +21,7 @@ import (
 	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rsa"
+	"errors"
 	"fmt"
 
 	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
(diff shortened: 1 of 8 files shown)


pkg/signature/algorithm_registry.go
```diff
@@ -222,3 +223,48 @@ func ParseSignatureAlgorithmFlag(flag string) (v1.PublicKeyDetails, error) {
...
+		switch pk.Curve {
+		case elliptic.P256():
+			return v1.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256, nil
+		case elliptic.P384():
+			return v1.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384, nil
+		case elliptic.P521():
+			return v1.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512, nil
+		}
+	case ed25519.PublicKey:
+		return v1.PublicKeyDetails_PKIX_ED25519_PH, nil
```
haydentherapper, line 257: To confirm, do we want the PH variant to be the default rather than the non-ph variant? Will this break anything?

Also, for Fulcio, there is no OID for ed25519-ph, or at least it's not a part of crypto/x509. In this case, would we just use the OID for ed25519? I think this came up before but I don't recall us deciding on if we are OK with this.

pkg/signature/algorithm_registry.go
```diff
@@ -222,3 +223,48 @@ func ParseSignatureAlgorithmFlag(flag string) (v1.PublicKeyDetails, error) {
...
+// GetDefaultPublicKeyDetails returns the default public key details for a given key.
+//
+// RSA 2048 => v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_2048_SHA256
+// RSA 3072 => v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_3072_SHA256
+// RSA 4096 => v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_4096_SHA256
+// ECDSA P256 => v1.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256
+// ECDSA P384 => v1.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384
+// ECDSA P521 => v1.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512
+// ED25519 => v1.PublicKeyDetails_PKIX_ED25519_PH
+func GetDefaultPublicKeyDetails(publicKey crypto.PublicKey) (v1.PublicKeyDetails, error) {
```
haydentherapper, line 236: would we want a variant of this function that allows us to pass options, if a client wanted to support PSS instead of PKCS1v1.5 instead? This might also be the answer to ed25519 vs -ph.

pkg/signature/algorithm_registry.go
ret2libc: > To confirm, do we want the PH variant to be the default rather than the non-ph variant? Will this break anything?

It might break cosign using ed25519 keys, because older blobs are signed with ed25519, but the new cosign versions, using this function, are going to use ed25519ph. This is relevant only for people using `--tlog-upload=false` (because they couldn't upload to rekor anyway, AFAIK).

Indeed, we discussed in the past having some logic in Cosign sign-blob to use ed25519 or ed25519ph based on `--tlog-upload`, while having a ed25519ph+ed25519 fallback validation during verify-blob, so that first cosign would try to validate with ed25519ph and if that fails it would try to validate with ed25519.

I think we could just always use ed25519ph for signing new blobs, while developing the ed25519ph+ed25519 fallback validation in the sigstore-go verification. What do you think? Are we still ok with this plan?

pkg/signature/algorithm_registry.go
ret2libc: One thing that might change if we do this is that ed25519ph needs to be enabled with the `WithED25519ph`, so by default with no options you select PureEd25519

pkg/signature/algorithm_registry.go
haydentherapper: I'm good with that solution.

haydentherapper (approved):
I like the suggestion to add Default to the method names. Otherwise this looks great.

haydentherapper:
Would you like me to merge as-is, or update the func names now with `Default`?

ret2libc:
> Would you like me to merge as-is, or update the func names now with `Default`?

I was planning on doing a bunch of changes tomorrow to the various PRs, but I can do this now!

haydentherapper:
No rush, just wanted to check!

haydentherapper (approved):

pkg/signature/algorithm_registry.go
```diff
@@ -222,3 +223,48 @@ func ParseSignatureAlgorithmFlag(flag string) (v1.PublicKeyDetails, error) {
...
+			return v1.PublicKeyDetails_PKIX_RSA_PKCS1V15_4096_SHA256, nil
+		}
+	case *ecdsa.PublicKey:
+		switch pk.Curve {
+		case elliptic.P256():
+			return v1.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256, nil
+		case elliptic.P384():
+			return v1.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384, nil
+		case elliptic.P521():
+			return v1.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512, nil
```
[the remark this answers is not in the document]
ret2libc: it is ECDSA P521 not P512 , see https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/P521_SHA512.pdf

PR #2024, opened 2025-03-11
Add tink package

cmurphy:
Copy in the tink package from Rekor v1 so that it can be used by other services.

Relates to https://github.com/sigstore/rekor-tiles/issues/9

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,9 +19,11 @@ require (
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966
 	github.com/stretchr/testify v1.10.0
 	github.com/theupdateframework/go-tuf v0.7.0
+	github.com/tink-crypto/tink-go/v2 v2.3.0
 	golang.org/x/crypto v0.35.0
 	golang.org/x/oauth2 v0.27.0
 	golang.org/x/term v0.29.0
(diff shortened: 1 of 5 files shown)


haydentherapper (commented):
Should we copy in the code for `GetPrimaryKey` which initializes a Tink decrypter given a KMS key, and `NewTinkSigner` to decrypt a keyset given a key? https://github.com/sigstore/timestamp-authority/blob/1abfce06d86c2e78094ec537ddebc92c4f9a051b/pkg/signer/tink.go#L53-L100 (and [tests](https://github.com/sigstore/timestamp-authority/blob/1abfce06d86c2e78094ec537ddebc92c4f9a051b/pkg/signer/tink_test.go#L43))

There's a bit of deviation between NewTinkSigner across Fulcio, Rekor and the TSA. The TSA implementation looks to be the most up to date.

cmurphy:
> Should we copy in the code for GetPrimaryKey which initializes a Tink decrypter given a KMS key, and NewTinkSigner to decrypt a keyset given a key?

This isn't ideal for what we're trying to use it for in rekor, `NewTinkSigner` returns a `crypto.Signer` when what we need is a `sigstore/sigstore/pkg/signature.Signer`

Issue #2042, opened 2025-04-04
Support for extracting signature algorithm from X509 certificate


loosebazooka:
As part of the algorithm registry work. Extract the `SignatureAlgorithm` from the x509 certificate and map the provided OID to sigstore types: https://github.com/golang/go/blob/master/src/crypto/x509/x509.go#L334-L347

PR #2043, opened 2025-04-04
Get algorithm details by OID

haydentherapper:
This will be used by Rekor to determine which hash function should be used for a given key algorithm. We already have a function to map PublicKeyDetails to AlgorithmDetails which works for keys, since Rekor v2 requests will require the user specify this value. However for certificates, we don't have this enum, so instead we'll use the OID from the certificate.

We add a few additional options to differentiate between different RSA key sizes, since RSA-PSS and RSA-PKCS#1v1.5 OIDs are the same for all key sizes, and between Ed25519 and Ed25519-ph.

Fixes https://github.com/sigstore/sigstore/issues/2042

### DIFF
--- a/pkg/signature/algorithm_registry.go
+++ b/pkg/signature/algorithm_registry.go
@@ -21,6 +21,7 @@ import (
 	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rsa"
+	"encoding/asn1"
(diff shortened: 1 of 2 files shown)


haydentherapper:
NVM, this is not working as intended. Got mixed up on what the signature algorithm is for.

PR #2068, opened 2025-04-22
Add a Name field to the TargetFile struct

bkabrda:
#### Summary
This PR is related to https://github.com/sigstore/cosign/issues/4098 - in order to fix that issue, the cosign code has to access the target name to see if it's one of the "fallbacks" or not (and treat it accordingly as explained in the linked issue).

I really wanted to keep this change contained to cosign codebase itself, but I just couldn't find a way to do this; I think adding this small PR here shouldn't hurt, as it is fully backwards compatible.

#### Release Note

Added a `Name` field to the `TargetFile` struct

#### Documentation

I believe this PR doesn't require any documentation update


bkabrda:
@cpanato Will do. I accidentally reset my Git signing config and I am struggling to get it to work again, but I should have it ready soon.

bkabrda:
@cpanato done, I guess the DCO check didn't register me uploading the new public key to my GH account, so it might need to be rerun now (?) The GH UI is already telling me that the commit is "Verified", so I think I'm all set on my side.

bkabrda:
@cpanato I think I forgot to force push, sorry. Should be all ok now.

haydentherapper (approved):
This TUF client is deprecated but this is fine for fixing the bug. The longer-term fix is to migrate to the TUF v2 client and TrustedRoot, but this is of course a larger fix.

Issue #2065, opened 2025-04-17
SignMessage accepts a context via WithContext, but the context isn't used.


ejahnGithub:
**Description**

https://github.com/sigstore/sigstore/blob/844f42dea5610211c3e479b26e18499d735d4faf/pkg/signature/kms/azure/signer.go#L88

```
func (a *SignerVerifier) SignMessage(message io.Reader, opts ...signature.SignOption) ([]byte, error) {
```
`SignMessage` takes WithContext, but context is not being used in the function.

bobcallaway:
fixed by #2070

PR #2070, opened 2025-04-25
Add context from opts to Azure signer

ejahnGithub:
#### Summary
per https://github.com/sigstore/sigstore/issues/2065

This pull request adds the context handling to the Azure SignerVerifier implementation. It ensures to use the context passing from opts for context-based configurations.

### DIFF
--- a/pkg/signature/kms/azure/signer.go
+++ b/pkg/signature/kms/azure/signer.go
@@ -87,7 +87,7 @@ func LoadSignerVerifier(defaultCtx context.Context, referenceStr string) (*Signe
 // All other options are ignored if specified.
 func (a *SignerVerifier) SignMessage(message io.Reader, opts ...signature.SignOption) ([]byte, error) {
 	var digest []byte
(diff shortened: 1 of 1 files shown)


pkg/signature/kms/azure/signer.go
```diff
@@ -87,12 +87,14 @@ func LoadSignerVerifier(defaultCtx context.Context, referenceStr string) (*Signe
 // All other options are ignored if specified.
 func (a *SignerVerifier) SignMessage(message io.Reader, opts ...signature.SignOption) ([]byte, error) {
 	var digest []byte
+	ctx := context.Background()
```
bdehamer, line 90: Since the `SignerVerifier` is still instantiated with a default context, should we use that as the default instead of `context.Background`?

bdehamer: Nevermind, @codysoyland beat me to it

codysoyland (approved):

Issue #2130, opened 2025-08-04
bug: invalid private key with localkms test


arthurus-rex:
**Description**

In attempting to run the [local KMS example](https://github.com/sigstore/sigstore/tree/main/test/cliplugin/localkms), I am experiencing a persistant `invalid private key PEM block` error.  The RSA 2048-bit private key was generated using OpenSSL (I tried both wrapped and unwrapped with `cosign import-key-pair`), and the error is the same for both `cosign sign` and `cosign sign-blob`.

I am still troubleshoting this myself, but putting this issue out here in case anyone has any insights.  This is the full error dump when running `cosign sign --key localkms:///path/to/my/test_key.pem $IMAGE`:
```
([]string) (len=3 cap=3) {
 (string) (len=21) "sigstore-kms-localkms",
 (string) (len=2) "v1",
 (string) (len=456) "{\"initOptions\":{\"ctxDeadline\":\"2025-08-04T17:48:46.608975735-04:00\",\"protocolVersion\":\"v1\",\"keyResourceID\":\"/path/to/my/test_key.pem\",\"hashFunc\":5,\"rpcOptions\":{\"keyVersion\":\"\",\"remoteVerification\":false}},\"methodName\":\"signMessage\",\"signMessage\":{\"signOptions\":{\"rpcOptions\":{\"ctxDeadline\":\"2025-08-04T17:48:46.608975735-04:00\",\"keyVersion\":\"\",\"remoteVerification\":false},\"messageOptions\":{\"digest\":null}}}}"
}
(*common.PluginArgs)(0xc00002a950)({
 InitOptions: (*common.InitOptions)(0xc000094fc0)({
  CtxDeadline: (*time.Time)(0xc000010588)(2025-08-04 17:48:46.608975735 -0400 EDT),
  ProtocolVersion: (string) (len=2) "v1",
  KeyResourceID: (string) (len=72) "/path/to/my/test_key.pem",
  HashFunc: (crypto.Hash) SHA-256,
  RPCOptions: (*common.RPCOptions)(0xc0000105a0)({
   CtxDeadline: (*time.Time)(<nil>),
   KeyVersion: (*string)(0xc00002aa60)(""),
   RemoteVerification: (*bool)(0xc00021c56e)(false)
  })
 }),
 MethodArgs: (*common.MethodArgs)(0xc000095040)({
  MethodName: (string) (len=11) "signMessage",
[...]

PR #2131, opened 2025-08-04
use more generic PEM parser within localkms example

bobcallaway:
fyi @arthurus-rex

Fixes: #2130


haydentherapper (approved):
Thanks for the quick fix!

PR #2194, opened 2025-11-04
cryptoutils: move goodkey validation to separate package

tonistiigi:
#### Summary

Moved `goodkey` based `ValidatePubKey` function to separate package.

While cryptoutils pkg provides a collection of locally defined
utility functions, it also provides a small wrapper around external
goodkey policy package from LetsEncrypt boulder.

This package is more opinionated and has a large external
dependency that it imports in. Therefore, it should be defined in
a separate package so that users who actually need this behavior
can import it directly and other users don't need to worry about
unwanted dependency.

This change reduces `sigstore-go/verifier` footprint by
```
 51 files changed, 3 insertions(+), 12923 deletions(-)
```
Removing `github.com/go-jose/go-jose/v4`, `github.com/letsencrypt/boulder`, `github.com/titanous/rocacheck`, `golang.org/x/crypto/ocsp`.

#### Release Note

The `ValidatePubKey` function has been moved from `pkg/cryptoutils` to `pkg/cryptoutils/goodkey` for better dependency control.

### DIFF
--- a/pkg/cryptoutils/goodkey/publickey_test.go
+++ b/pkg/cryptoutils/goodkey/publickey_test.go
@@ -95,7 +95,7 @@ func TestValidatePubKeyEcdsa(t *testing.T) {
 			t.Errorf("expected error when validating public key")
 		}
 	}
-	// Fails with smalller curve
+	// Fails with smaller curve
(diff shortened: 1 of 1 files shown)


haydentherapper:
Just need to fix lint errors and looks like you need to run go mod tidy for some of the inner modules

PR #2326, opened 2026-05-12
add functional options to DSSE to improve memory usage, validation

bobcallaway:
- Bump go-securesystemslib from v0.9.1 to v0.11.0                                                                                                                                                       
- Replace all Verify calls with VerifyAndDecode, which returns the decoded payload and avoids a redundant base64 decode for callers that need it
- Introduce dsse.Option functional options for verifier construction:                                                                                                                                   
  - WithDecodedPayload(*[]byte) — writes the decoded envelope payload into the provided pointer after successful verification, enabling zero-copy access to the payload                                 
  - WithExpectedPayloadType(string) — rejects envelopes whose payload type doesn't match the expected value before performing signature verification                                                    
- WrapSignerVerifier and WrapMultiSignerVerifier now automatically enforce payload type matching using the payloadType they already accept at construction                                              
- New WrapMultiVerifierWithOpts and WrapMultiSignerVerifierWithOpts take a []signature.Verifier / []signature.SignerVerifier slice instead of variadic args, to allow the ...Option variadic                 
                                                                                                                                                                                                          
### Test plan                                                                                                                                                                                               
                                                                                                                                                                                                        
- Existing round-trip tests pass (single + multi)                                                                                                                                                       
- TestRoundTripWithDecodedPayload — decoded payload matches original data
- TestMultiRoundTripWithDecodedPayload — same for multi verifier                                                                                                                                        
- TestExpectedPayloadTypeMatch — matching type passes verification                                                                                                                                      
- TestExpectedPayloadTypeMismatch — wrong type fails verification                                                                                                                                       
- TestSignerVerifierEnforcesPayloadType — auto-enforced via WrapSignerVerifier                                                                                                                           - TestMultiExpectedPayloadTypeMismatch — wrong type fails for multi verifier                                                                                                                            
- TestMultiSignerVerifierEnforcesPayloadType — auto-enforced via WrapMultiSignerVerifier


pkg/signature/dsse/options.go
```diff
@@ -0,0 +1,53 @@
...
+	return func(c *wrapConfig) {
+		c.decodedPayload = p
+	}
+}
+
+// WithExpectedPayloadType returns an Option that causes the verifier to
+// check the envelope's payload type against the expected value before
+// verifying. If the types do not match, verification fails immediately.
+// When this option is not set, any payload type is accepted.
+func WithExpectedPayloadType(t string) Option {
```
Hayden-IO, line 41: To confirm, is this a config option because verification didn't compare payload types before? But it is for wrapping the SignerVerifier. Should this be a required parameter for verify functions? We're making breaking function changes in other places already.

pkg/signature/dsse/multidsse.go
```diff
@@ -86,13 +87,13 @@ func (wL *wrappedMultiSigner) SignMessage(r io.Reader, _ ...signature.SignOption
...
-	threshold   int
-	payloadType string
+	vLAdapters []dsse.Verifier
+	threshold  int
+	cfg        wrapConfig
 }
 
 // WrapMultiVerifier returns a signature.Verifier that uses the DSSE encoding format
-func WrapMultiVerifier(payloadType string, threshold int, vL ...signature.Verifier) signature.Verifier {
+func WrapMultiVerifier(payloadType string, threshold int, vL []signature.Verifier, opts ...Option) signature.Verifier {
```
Hayden-IO: We can't introduce a breaking API change without bumping to a v2 - can you keep the existing method signatures around as well but mark them as deprecated?

pkg/signature/dsse/options.go
bobcallaway: correct: https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#protocol

payload type should be agreed upon OOB

pkg/signature/dsse/multidsse.go
bobcallaway: is this strictly breaking (given the move from a variadic to an explicit option? I thought this "might just work" (TM)

Hayden-IO: The breakage is because the function can be called with no parameters (verified with https://go.dev/play/p/JtCbWF98BaT), but maybe I'm being unnecessarily paranoid because calling `WrapMultiVerifier` with an empty `vL` is arguably a bug as there should be at least one verifier.

bobcallaway: we can back out the change on the multiverifier if you're worried that's not ideal - Sigstore probably only cares about singlely signed envelopes

Hayden-IO: Looks like this is a breaking change because you can't pass variadic arguments using `...` to an array parameter. https://go.dev/play/p/QA8CV3rlk-E

Found an example of this function be called [here](https://github.com/deepin-community/golang-github-sylabs-sif/blob/175433a68dba35f157a99c0e626536e0ddd1c06d/pkg/integrity/dsse.go#L100).

Can we keep the existing function, and add a second with another name like WrapMultiVerifierWithOpts and deprecate the other?

Hayden-IO (approved):
thanks!

PR #2332, opened 2026-05-27
Extend PEM private key unmarshalling to support legacy format

Hayden-IO:
This allows decryption of standard PEM-encrypted private keys directly during parsing. This will be used by other Sigstore libraries and services to avoid pulling in 3rd party dependencies to handle private key decryption.


Hayden-IO:
I'm adding this so we can drop the dependency on smallstep/crypto#pemutil.Read, which will reduce the dep tree.

steiza (approved):
Looks reasonable to me. I'm a little out of my depth here - I think this is for Rekor `pkg/signer/file.go`? I'm not familiar with how that function is used, but these changes seem like they'll work.

Hayden-IO:
Yea, this is for rekor, fulcio and the TSA, all of which can load an encrypted signing key from disk.

PR #2333, opened 2026-05-29
Support standard PKCS#8 encrypted private key decryption

Hayden-IO:
Extend UnmarshalPEMToPrivateKey to support standard PKCS#8 encrypted private keys ("ENCRYPTED PRIVATE KEY"). This was missed with the previous change, which supported RFC 1423 encrypted keys but not RFC 7468. We do have to depend on a 3p dependency, however this dependency only uses the std library.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/theupdateframework/go-tuf v0.7.0
 	github.com/tink-crypto/tink-go/v2 v2.6.0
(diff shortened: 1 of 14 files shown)


steiza (approved):
I was trying to figure out if there was a better way to store private keys encrypted on disk than `PKCS #8`, but it seems like this is what things like the timestamp authority expect / support today.

Hayden-IO:
For production, the recommendation is to use Tink, with a KMS-backed key-encryption-key that wraps a private key that's decrypted in memory. But we continue to support on-disk keys, with simple encryption using OpenSSL or similar (which is not side-channel resistant, but it's better than nothing).

PR #2353, opened 2026-06-09
Support ML-DSA signing and verification

### LINKED ISSUE #2352: ML-DSA support for signer/verifier and certificates
**Description**

Go 1.27 will land support for ML-DSA in crypto/mldsa, along with crypto/x509 to support ML-DSA certificates and PKIX encodings. This issue tracks the changes to add ML-DSA support to sigstore/sigstore.

We'd like to introduce APIs to be used by other Sigstore services and Go SDKs. We can leverage filippo.io/mldsa in the short-term, however it will be a breaking change for any exported APIs that accept or return types defined in filippo.io/mldsa when we switch to crypto/mldsa.

In the short-term, I propose we introduce some duplication within the Go projects - we'll have a PR up with what other Sigstore projects can use, and then we'll commit the changes once 1.27 lands.

Hayden-IO:
Fixes https://github.com/sigstore/sigstore/issues/2352


pkg/signature/mldsa.go
```diff
@@ -0,0 +1,192 @@
...
+// supported as ML-DSA handles its own internal message processing.
+//
+// All options are ignored.
+func (m MLDSASigner) SignMessage(message io.Reader, _ ...SignOption) ([]byte, error) {
+	messageBytes, _, err := ComputeDigestForSigning(message, crypto.Hash(0), mldsaSupportedHashFuncs)
+	if err != nil {
+		return nil, err
+	}
+
+	return m.priv.SignDeterministic(messageBytes, nil)
```
Hayden-IO: Note: Change to the non-deterministic variant

pkg/signature/sv_test.go
Hayden-IO, line 1: Add more tests

Hayden-IO:
Looking for an early review - this will move out of draft once I bump the repo to Go 1.27.0 and update filippo.io/mldsa package usage to the std lib.

pkg/cryptoutils/publickey.go
```diff
@@ -112,6 +115,10 @@ func EqualKeys(first, second crypto.PublicKey) error {
 		if !pub.Equal(second) {
 			return errors.New(genErrMsg(first, second, "ed25519"))
 		}
+	case *mldsa.PublicKey:
+		if !pub.Equal(second) {
+			return errors.New(genErrMsg(first, second, "mldsa"))
+		}
```
jku, line 121: TestEqualKeys does not do much but you could still add a case for this

pkg/signature/signerverifier.go
```diff
@@ -68,6 +69,8 @@ func LoadSignerVerifierWithOpts(privateKey crypto.PrivateKey, opts ...LoadOption
 			return LoadED25519phSignerVerifier(pk)
 		}
 		return LoadED25519SignerVerifier(pk)
+	case *mldsa.PrivateKey:
+		return LoadMLDSASignerVerifier(pk)
 	}
 	return nil, errors.New("unsupported public key type")
```
jku, line 75: unrelated nit: 
```suggestion
	return nil, errors.New("unsupported private key type")
```

pkg/signature/kms/aws/client.go
```diff
@@ -260,9 +260,9 @@ func (a *awsClient) createKey(ctx context.Context, algorithm string) (crypto.Pub
 	usage := types.KeyUsageTypeSignVerify
 	description := "Created by Sigstore"
 	key, err := a.client.CreateKey(ctx, &kms.CreateKeyInput{
-		CustomerMasterKeySpec: types.CustomerMasterKeySpec(algorithm),
-		KeyUsage:              usage,
-		Description:           &description,
+		KeySpec:     types.KeySpec(algorithm),
+		KeyUsage:    usage,
+		Description: &description,
```
jku, line 265: this is just a AWS API update right?

jku (approved):
This seems complete to me.

You could still add the ml-dsa equivalent of `TestECDSAPublicKeyPEMRoundtrip`.

pkg/signature/algorithm_registry.go
```diff
@@ -140,6 +156,16 @@ func (a AlgorithmDetails) checkKey(pubKey crypto.PublicKey) (bool, error) {
...
+	case MLDSA:
+		mldsaKey, ok := pubKey.(*mldsa.PublicKey)
+		if !ok {
+			return false, nil
+		}
+		params, err := a.GetMLDSAParameters()
+		if err != nil {
+			return false, err
+		}
+		return mldsaKey.Parameters() == params, nil
```
jku, line 168: this seems to panic on malformed (like uninitialized or nil) keys. Same in GetDefaultPublicKeyDetails

I wonder if it's worth writing a `SafeMldsaParameters()` wrapper that does not panic?

pkg/signature/mldsa.go
```diff
@@ -0,0 +1,191 @@
...
+	return &MLDSASigner{
+		priv: priv,
+	}, nil
+}
+
+// SignMessage signs the provided message. Passing the WithDigest option is not
+// supported as ML-DSA handles its own internal message processing.
+//
+// All options are ignored.
+func (m MLDSASigner) SignMessage(message io.Reader, _ ...SignOption) ([]byte, error) {
```
jku, line 53: should likely guard against `m.priv == nil` like other sign implementations

pkg/cryptoutils/publickey.go
jku: this code also panics on nil keys

pkg/signature/mldsa.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// LoadMLDSASignerVerifier creates a combined signer and verifier. This is
+// a convenience object that simply wraps an instance of MLDSASigner and MLDSAVerifier.
+func LoadMLDSASignerVerifier(priv *mldsa.PrivateKey) (*MLDSASignerVerifier, error) {
+	signer, err := LoadMLDSASigner(priv)
+	if err != nil {
+		return nil, fmt.Errorf("initializing signer: %w", err)
+	}
+	pub, ok := priv.Public().(*mldsa.PublicKey)
```
jku, line 149: nit: `priv.PublicKey()` should work here

pkg/signature/mldsa.go
```diff
@@ -0,0 +1,191 @@
...
+// PublicKey returns the public key that can be used to verify signatures created by
+// this signer. As this value is held in memory, all options provided in arguments
+// to this method are ignored.
+func (m MLDSASigner) PublicKey(_ ...PublicKeyOption) (crypto.PublicKey, error) {
+	return m.Public(), nil
+}
+
+// Sign computes the signature for the specified message; the first and third arguments to this
+// function are ignored as they are not used by the ML-DSA algorithm (using deterministic signing).
+func (m MLDSASigner) Sign(_ io.Reader, message []byte, _ crypto.SignerOpts) ([]byte, error) {
```
jku, line 81: Re-reading this:
* I believe "deterministic" is incorrect: signing is still randomized AFAIK
* the comment could explain that we use _"Pure ML-DSA only with an empty context for consistency across software, KMS, and hardware backends"_ (with maybe a hint that callers can enforce their own domain separation at payload level) -- otherwise dropping options feels arbitrary and surprising
* could also add an option guard to make sure context or MLDSAMu are not used

jku (commented):
Did another pass:
* some comments on making sign() a little more clear on how it's special compared upstream (go crypto) ML-DSA signing
* nil and empty keys seems to panic the go mldsa module by design. Unsure if we should do something about it -- this code might be used against public keys we don't really know...

jku:
I believe this can be closed: #2415, #2416, #2417 handle all of it

PR #2355, opened 2026-06-11
Fix Azure KMS support for RSA signatures

### LINKED ISSUE #2354: Azure KMS not supporting RSA signatures
**Description**

In https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/azure/signer.go#L112 the resulting signature is unconditionally wrapped in the ASN.1 EC signature format, because Azure returns them as plain `r || s` and go expects ASN.1.

This of course does not work for RSA keys, and the signature will fail when using RSA keys. This also causes cosign to fail, like described in https://github.com/sigstore/cosign/issues/4915

I will send in a PR shortly to enable support for RSA keys in the Azure KMS.

**Version**

cosign 3.0.6

boomer41:
#### Summary

Fixes #2354.

RSA signatures have been internally half-ish-supported by the Azure KMS provider,
but this circumstance is rather undocumented and will throw random errors when actually trying to use it.
See https://github.com/sigstore/cosign/issues/4915 for this.

This PR adds proper support for RSA signatures.

#### Release Note

- Azure KMS provider now supports RSA keys

#### Documentation

No docs necessary, bugfix.

### DIFF
--- a/pkg/signature/kms/azure/client.go
+++ b/pkg/signature/kms/azure/client.go
@@ -269,41 +269,43 @@ func (a *azureVaultClient) createKey(ctx context.Context) (crypto.PublicKey, err
 	return a.public(ctx)
 }
 
-func (a *azureVaultClient) getKeyVaultHashFunc(ctx context.Context) (crypto.Hash, azkeys.SignatureAlgorithm, error) {
+// 3rd return value specifies whether the resulting signature is an EC signature,
+// thus requiring conversion between the signature formats.
+func (a *azureVaultClient) getKeyVaultHashFunc(ctx context.Context) (crypto.Hash, azkeys.SignatureAlgorithm, bool, error) {
 	publicKey, err := a.public(ctx)
 	if err != nil {
-		return 0, "", fmt.Errorf("failed to get public key: %w", err)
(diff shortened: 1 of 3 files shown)


Hayden-IO:
Can you please rebase?

boomer41:
@Hayden-IO done. :)

Hayden-IO (dismissed):
I unfortunately don't have a way to test this, but this looks alright to me. Have you gotten a chance to test using cosign with this change?

boomer41:
Yes, I have in fact built an image here containing the changes for my internal use: https://hub.docker.com/r/boomer41/cosign

With these changes, the error mentioned in the issue goes away and a signature is attached.

boomer41:
If you do need a Key vault with a test key attached, I can provide you with necessary test credentials via private channel.

Hayden-IO:
Fantastic! It's alright, it working is sufficient, thank you though.

boomer41:
I will push another commit fixing the linter, as it was just run now.

boomer41:
@Hayden-IO it seems you need to review again unfortunately.

Hayden-IO (approved):

PR #2369, opened 2026-07-10
Deprecate TUF client, ValidatePubKey

Hayden-IO:
sigstore-go is the only supported Sigstore Go TUF client now. To remove a large dependency, letsencrypt/boulder, we'll deprecate ValidatePubKey, which really isn't necessary except for double checking imported keys aren't weak. As we're moving towards PQC, this shouldn't be needed anyways.


go.mod
```diff
@@ -24,19 +23,17 @@ require (
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
+	github.com/nxadm/tail v1.4.11 // indirect
```
Hayden-IO, line 26: I don't know why this just got pulled in, must be go mod tidy wasn't run recently.

pkg/fulcioroots/fulcioroots.go
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 
 // Package fulcioroots fetches Fulcio root and intermediate certificates from TUF metadata
+//
+// Deprecated: Use https://pkg.go.dev/github.com/sigstore/sigstore-go@main/pkg/tuf
```
steiza, line 18: Just FYI, this package is used in a Cosign CLI subcommand that isn't wired up anymore (I guess we should delete it) as well as Cosign e2e tests.

pkg/cryptoutils/goodkey/publickey.go
```diff
@@ -13,68 +13,20 @@
...
-	"errors"
-	"fmt"
 
-	"github.com/letsencrypt/boulder/goodkey"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
 
 // ValidatePubKey validates the parameters of an RSA, ECDSA, or ED25519 public key.
 func ValidatePubKey(pub crypto.PublicKey) error {
-	// goodkey policy enforces:
```
steiza, line 33: This is used in Cosign, but we can just use `github.com/sigstore/sigstore/pkg/cryptoutils` directly.

steiza (approved):
LGTM! Two minor comments.

pkg/fulcioroots/fulcioroots.go
Hayden-IO: Yep, I started to remove this and the old TUF client from Cosign and realized that this is more or less already done with the PR removing the old signing and verification logic.

pkg/cryptoutils/goodkey/publickey.go
Hayden-IO: I wish I had just removed boulder as a dependency a bit ago. Because it's a large dependency and its license is different, there was a request to have it moved to its own package. We'll just drop its usage in Cosign.

PR #2378, opened 2026-07-14
Validate ed25519 public key length in verifier constructors

sueun-dev:
`LoadED25519Signer` rejects a private key that isn't `ed25519.PrivateKeySize`, with the comment "check this to avoid panic and throw error gracefully". `LoadED25519Verifier` and `LoadED25519phVerifier` only check for a nil key, so a non-nil `ed25519.PublicKey` of the wrong length is accepted and the panic just moves to `VerifySignature`: crypto/ed25519's `Verify` / `VerifyWithOptions` call `panic("ed25519: bad public key length: N")`.

`LoadVerifier`, `LoadVerifierWithOpts` and `LoadUnsafeVerifier` all dispatch ed25519 keys straight to these constructors. A caller that builds a key from raw bytes, rather than going through `cryptoutils.UnmarshalPEMToPublicKey` (which always yields a 32-byte key), gets a panic at verify time instead of a load error.

This adds the same length check the signer already has, so the constructors return `invalid size for ED25519 key` up front.

Checked:
- `go test ./pkg/signature/...` passes
- the new `TestED25519VerifierRejectsWrongKeySize` / `TestED25519phVerifierRejectsWrongKeySize` fail before the change (loader returns a nil error for wrong-length keys) and pass after

### DIFF
--- a/pkg/signature/ed25519.go
+++ b/pkg/signature/ed25519.go
@@ -101,6 +101,11 @@ func LoadED25519Verifier(pub ed25519.PublicKey) (*ED25519Verifier, error) {
 		return nil, errors.New("invalid ED25519 public key specified")
 	}
 
+	// check this to avoid a panic in ed25519.Verify and throw an error gracefully
+	if len(pub) != ed25519.PublicKeySize {
+		return nil, errors.New("invalid size for ED25519 key")
+	}
+
 	return &ED25519Verifier{
 		publicKey: pub,
 	}, nil
--- a/pkg/signature/ed25519.go
+++ b/pkg/signature/ed25519.go
@@ -101,7 +101,6 @@ func LoadED25519Verifier(pub ed25519.PublicKey) (*ED25519Verifier, error) {
 		return nil, errors.New("invalid ED25519 public key specified")
 	}
 
-	// check this to avoid a panic in ed25519.Verify and throw an error gracefully
 	if len(pub) != ed25519.PublicKeySize {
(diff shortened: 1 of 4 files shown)


pkg/signature/ed25519.go
```diff
@@ -101,6 +101,11 @@ func LoadED25519Verifier(pub ed25519.PublicKey) (*ED25519Verifier, error) {
 		return nil, errors.New("invalid ED25519 public key specified")
 	}
 
+	// check this to avoid a panic in ed25519.Verify and throw an error gracefully
```
Hayden-IO: Comment can be removed, the code is self documenting.

pkg/signature/ed25519_test.go
```diff
@@ -86,3 +86,18 @@ func TestED25519Verifier(t *testing.T) {
 	}
 	assertPublicKeyIsx509Marshalable(t, pub)
 }
+
+// TestED25519VerifierRejectsWrongKeySize ensures LoadED25519Verifier validates
```
Hayden-IO: Same here, the comment isn’t adding additional context.

Hayden-IO (commented):
I’m not sure how a key would have been generated that’s the wrong length, but I think this is a fine defensive check.

Hayden-IO:
Just bumping for comments

Hayden-IO (approved):

bobcallaway (approved):

PR #2383, opened 2026-07-28
Add BrowserOpener field to InteractiveIDTokenGetter

wlynch:
#### Summary

This adds a new BrowserOpener field to allow implementations to customize the browser opening experience. Existing behavior remains the same.

We want to use this in gitsign in order to allow users to provide custom browser opening commands, which allows users to force Chrome to open with certain profiles rather than relying on the default behavior of browser of "last window that happens to be open".

#### Release Note

* Adds ability for client libraries to provide custom URL openers.

#### Documentation

n/a


cpanato (approved):
thanks!

PR #2394, opened 2026-08-20
Bump Go version to 1.27.0

Hayden-IO:
This is a prereq to #2352, to use crypto/mldsa.

This also fixes new linter failures around manual manipulation of ECDSA key parameters. There is now a method to parse a private key, and an invalid key no longer causes a panic.


loosebazooka:
this has been rebased on main and updated for other changes in #2415

PR #2396, opened 2026-08-24
Add retryable HTTP transport

crazy-max:
#### Summary

This adds a shared no-dependency HTTP retry package to `github.com/sigstore/sigstore` so Sigstore clients and services have a common retry transport instead of each package choosing its own third-party helper or reimplementing retry policy locally. This is related to the sigstore-go work in https://github.com/sigstore/sigstore-go/pull/671, where default TUF downloads needed retry behavior and `github.com/hashicorp/go-retryablehttp` was not the right dependency tradeoff for this codebase.

The API keeps normal `net/http` integration as the main contract. Callers can create a retrying client with `httpretry.NewClient()` or wrap an existing base transport with `httpretry.NewTransport(...)`, and the implementation remains an `http.RoundTripper` rather than a separate request execution framework. The default policy retries transient request errors, HTTP 429, status code 0, and HTTP 5xx responses except 501, while avoiding retries for context cancellation, context deadlines, TLS certificate failures, invalid protocol or header configuration, and redirect-limit failures.

The default backoff uses capped exponential delay with jitter and honors `Retry-After` for HTTP 429 and 503 responses. The transport also closes failed response bodies before retrying and avoids retrying requests with bodies that cannot be replayed, which prevents callers from accidentally resending partial or empty request data. The scope is intentionally limited to the reusable retry behavior Sigstore callers need, without adding logging hooks, custom request wrapper types, a broad client framework, or new dependencies.

#### Release Note

A shared HTTP retry package was added with a retrying transport, default retry policy, jittered backoff, `Retry-After` handling, and request replay safeguards.

#### Documentation

No documentation update is required. Callers can use this utility directly from the Go package documentation by constructing a retrying client with `httpretry.NewClient()` or wrapping an existing transport with `httpretry.NewTransport(...)`.

cc @Hayden-IO

### DIFF
--- a/pkg/httpretry/client.go
+++ b/pkg/httpretry/client.go
@@ -1,0 +1,284 @@
+//
+// Copyright 2026 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package httpretry
+
+import (
+	"context"
+	"crypto/x509"
+	"errors"
+	"io"
+	"math/rand/v2"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	// DefaultMaxRetries is the default number of times to retry a failed request.
+	DefaultMaxRetries = 4
+	// DefaultWaitMin is the default minimum time to wait between attempts.
+	DefaultWaitMin = time.Second
+	// DefaultWaitMax is the default maximum time to wait between attempts.
+	DefaultWaitMax = 30 * time.Second
+)
+
+const drainBodyLimit = 2 * 1024 * 1024
+
+// CheckRetry decides whether a response or error should be retried.
+type CheckRetry func(req *http.Request, resp *http.Response, err error) bool
+
+// Backoff decides how long to wait before the next attempt.
+type Backoff func(waitMin, waitMax time.Duration, attempt int, resp *http.Response) time.Duration
+
+// Transport is an HTTP transport that retries transient failures. A Transport
+// must not be modified after first use.
+type Transport struct {
+	// Base is the underlying transport used for each attempt. If nil,
+	// http.DefaultTransport is used.
+	Base http.RoundTripper
+	// MaxRetries is the number of times a failed request is retried.
+	MaxRetries int
+	// WaitMin is the minimum backoff between attempts.
+	WaitMin time.Duration
+	// WaitMax is the maximum backoff between attempts.
+	WaitMax time.Duration
+	// CheckRetry decides which responses and errors are retried. If nil,
+	// DefaultRetryPolicy is used.
+	CheckRetry CheckRetry
+	// Backoff decides how long to wait between retries. If nil, DefaultBackoff
+	// is used.
+	Backoff Backoff
+}
+
+// NewClient returns an HTTP client with a retrying transport.
+func NewClient() *http.Client {
+	return &http.Client{
+		Transport: NewTransport(nil),
+	}
+}
+
+// NewTransport returns a transport with the default retry settings.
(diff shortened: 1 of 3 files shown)


Hayden-IO:
Overall this looks good! This is pretty similar to what I had proposed in https://github.com/sigstore/rekor/pull/293, though I think this approach looks cleaner. I asked Bob, one of the other service maintainers, to take a pass as well.

crazy-max:
fixed lint issues

pkg/httpretry/client.go
```diff
@@ -0,0 +1,293 @@
...
+	default:
+		return false
+	}
+}
+
+// DefaultBackoff returns an exponential backoff delay with jitter, capped
+// between min and max. It honors Retry-After for HTTP 429 and 503 responses
+// when present.
+func DefaultBackoff(waitMin, waitMax time.Duration, attempt int, resp *http.Response) time.Duration {
+	if retryAfter := retryAfterDelay(resp); retryAfter > 0 {
```
bobcallaway: if a server returns a `Retry-After` > `waitMax, it looks like we ignore the requested maximum here?

pkg/httpretry/client.go
```diff
@@ -0,0 +1,293 @@
...
+	// is used.
+	Backoff Backoff
+}
+
+// NewClient returns an HTTP client with a retrying transport.
+func NewClient() *http.Client {
+	return &http.Client{
+		Transport: NewTransport(nil),
+	}
+}
```
bobcallaway:
```suggestion
func NewClient() *http.Client {
    base := http.DefaultTransport.(*http.Transport).Clone()
	return &http.Client{
		Transport: NewTransport(base),
	}
}
```

pkg/httpretry/client.go
```diff
@@ -0,0 +1,293 @@
...
+
+func closeIdleConnections(transport http.RoundTripper) {
+	type closeIdler interface {
+		CloseIdleConnections()
+	}
+	if closer, ok := transport.(closeIdler); ok {
+		closer.CloseIdleConnections()
+	}
+}
```
bobcallaway:
```suggestion
}

// CloseIdleConnections closes idle connections held by the underlying transport.
  func (t *Transport) CloseIdleConnections() {
      closeIdleConnections(t.base)
  }
```

pkg/httpretry/client.go
bobcallaway: this would give each client its own pool

pkg/httpretry/client.go
```diff
@@ -0,0 +1,293 @@
...
+func sleep(ctx context.Context, delay time.Duration, transport http.RoundTripper) error {
+	if delay <= 0 {
+		return ctx.Err()
+	}
+	timer := time.NewTimer(delay)
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
+		closeIdleConnections(transport)
```
bobcallaway: why would you close idle connections just because this particular request context is complete?

pkg/httpretry/client.go
```diff
@@ -0,0 +1,293 @@
...
+// NewClient returns an HTTP client with a retrying transport.
+func NewClient() *http.Client {
+	return &http.Client{
+		Transport: NewTransport(nil),
+	}
+}
+
+// NewTransport returns a transport with the default retry settings.
+func NewTransport(base http.RoundTripper) *Transport {
+	return &Transport{
```
bobcallaway:
```suggestion
func NewTransport(base http.RoundTripper) *Transport {
	if base == nil {
	    base = http.DefaultTransport.(*http.Transport).Clone()
	}
	return &Transport{
```

bobcallaway: if no transport is given, i think cloning the default one is the correct move rather than reusing it

pkg/httpretry/client.go
bobcallaway: if `Retry-After` > `waitMax`, wouldn't you want `waitMax` to take precedence?

pkg/httpretry/client.go
crazy-max: hum yeah closing idle connections from one request context cancellation is too broad for a shared transport. I will remove that and make the retry sleep just returns the request context error

pkg/httpretry/client.go
crazy-max: yes makes sense. Will change `DefaultBackoff` so `waitMax` still bounds the returned delay even when the server sends `Retry-After`.

bobcallaway (approved):

thaJeztah:
@Hayden-IO PTAL 🤗

PR #2410, opened 2026-09-01
Fix Azure KMS ECDSA signature verification (r||s ordering + padding)

### LINKED ISSUE #2409: Azure KMS: ECDSA signature verification always fails (`r||r` instead of `r||s`)
## Version

`github.com/sigstore/sigstore/pkg/signature/kms/azure` present since #2355
(e.g. v1.10.9), still present on main.

## Description

`SignerVerifier.VerifySignature` in the Azure KMS provider (`pkg/signature/kms/azure`)
reconstructs the raw EC signature incorrectly: when converting the ASN.1
signature back to the `r||s` form that Azure Key Vault's `Verify` operation
expects, it uses `r` for **both** halves instead of `r` and `s`.

`pkg/signature/kms/azure/signer.go`, `VerifySignature`:

```go
rBytes := r.Bytes()
sBytes := r.Bytes()   // BUG: should be s.Bytes()

sigBytes = make([]byte, 0, len(rBytes)+len(sBytes))
sigBytes = append(sigBytes, rBytes...)
sigBytes = append(sigBytes, sBytes...)
```

Because the second half is `r` (not `s`), the reconstructed signature is never
valid, so every ECDSA verification through this provider fails. Callers see:

```
could not verify envelope: accepted signatures do not match threshold, Found: 0, Expected 1
```

## Impact

Affects all EC keys (P-256/P-384/P-521) the `signatureRequiresECWrapping`
branch only runs for EC. RSA is unaffected.

Affects any consumer that verifies via the KMS `SignerVerifier` (as opposed to
fetching the public key and verifying locally). Notably this breaks Sigstore
bundle (new format) verification with an Azure KMS EC key.

Downstream reports:

Kyverno: `ImageValidatingPolicy` cosign key/KMS verification fails on v1.19.0
for signatures stored as v0.3 sigstore bundle referrers
(kyverno/kyverno#17363).

Related family: sigstore/cosign#4915 (RSA), which motivated PR #2355.

## How this was discovered and validated

This surfaced downstream in Kyverno (kyverno/kyverno#17363): `ImageValidatingPolicy`
with a cosign Azure KMS EC key and `insecureIgnoreTlog: true`, verifying an
image whose signature is a v0.3 sigstore bundle referrer (dsse-envelope),
failed only inside the Kyverno admission controller with `Found: 0`, while
the cosign CLI verified the same image successfully.

The CLI-vs-controller split is the key diagnostic and points straight at this
bug:

`cosign verify --key azurekms://…` fetches the public key and verifies
locally (standard ECDSA path) it never calls the KMS provider's
`VerifySignature`, so it is unaffected.

Kyverno passes the Azure `SignerVerifier` as `CheckOpts.SigVerifier`, so
sigstore-go's bundle/DSSE verification calls
`SignerVerifier.VerifySignature` → the buggy `r||r` reconstruction → `Found: 0`.

Version window confirms the regression precisely:

Kyverno v1.18.1 vendors `pkg/signature/kms/azure` v1.10.5 (correct
`r||s`) → EC verification works.

Kyverno v1.19.0 vendors v1.10.9 (post-#2355, `r||r`) → EC verification
fails.

End-to-end validation of the fix: rebuilding Kyverno v1.19.0 with a replace
pointing `pkg/signature/kms/azure` at a locally patched module (the one-line
`s.Bytes()` fix + fixed-width padding) and deploying it to a live AKS cluster
made the previously-failing Azure EC signature verify successfully, while
an unsigned image was still correctly denied... i.e. the fix restores
correct verification without weakening enforcement.

## Regression

Introduced by #2355 ("Fix Azure KMS support for RSA signatures"). That change
gated the EC signature (un)wrapping behind `signatureRequiresECWrapping` so RSA
would no longer be ASN.1-wrapped; while restructuring the EC unwrap logic into
the new conditional, the `s.Bytes()` call became `r.Bytes()`.

Comparison:

Before #2355 (e.g. v1.10.5): `... append(rBytes...) ; append(sBytes...)` correct `r||s`.

After #2355 (current main / v1.10.9): second half is `r` - `r||r`.

The typo is still present on main
(`pkg/signature/kms/azure/signer.go`, `VerifySignature`).

## Steps to reproduce

The existing unit tests use a no-op mock (`testKVClient.Verify` returns success
unconditionally), so they never exercise real signature bytes and cannot catch
this. A hermetic reproduction using real crypto:

Create a `kvClient` mock whose `Sign` returns a raw `r||s` ECDSA signature and
whose `Verify` performs a real `ecdsa.Verify` on the raw `r||s` it receives
(mirroring Azure Key Vault).

`SignMessage` a payload, then `VerifySignature` the result.

`VerifySignature` returns an error (failed vault verification) even though
the signature is valid.

Real-world: cosign sign/verify a v0.3 bundle with an Azure Key Vault EC key
and verify it through the KMS `SignerVerifier` path (e.g. via Kyverno) → `Found: 0`.

## Additional (secondary) issue: leading-zero padding

Even after fixing the typo, `big.Int.Bytes()` strips leading zero bytes, so when
`r` or `s` has a high zero byte (~1/256 each) the reconstructed `r||s` is shorter
than `2 * coordinateSize`, which Azure Key Vault also rejects. This causes
intermittent verification failures. The coordinates should be left-padded to
the curve's byte width (e.g. `FillBytes`).

## Suggested fix

```go
pub, err := a.client.public(a.defaultCtx)
if err != nil {
    return fmt.Errorf("getting public key: %w", err)
}
ecPub, ok := pub.(*ecdsa.PublicKey)
if !ok {
    return fmt.Errorf("expected ECDSA public key, got %T", pub)
}
keyBytes := (ecPub.Curve.Params().BitSize + 7) / 8

sigBytes = make([]byte, 2*keyBytes)
r.FillBytes(sigBytes[:keyBytes])
s.FillBytes(sigBytes[keyBytes:])
```

This fixes both the r/s ordering and the fixed-width padding. A real-crypto
round-trip test (covering P-256/P-384/P-521 and RSA-2048/3072/4096) plus a
deterministic leading-zero encoding test can guard against regressions.

I have a fix + tests ready and can open a PR.

hugolevino:
Closes #2409

VerifySignature reconstructed the raw EC signature using r for both halves (r||r) instead of r||s, so every ECDSA verification through the Azure KMS SignerVerifier failed. Additionally, big.Int.Bytes() strips leading zero bytes, producing an incorrectly sized r||s that Azure Key Vault rejects.

Use the ECDSA public key curve size to left-pad both coordinates with FillBytes, fixing both the ordering and fixed-width padding. Add a real-crypto round-trip test covering P-256/P-384/P-521 and RSA-2048/3072/4096.

Regression introduced by #2355.

#### Summary

`SignerVerifier.VerifySignature` in the Azure KMS provider (`pkg/signature/kms/azure/signer.go`) reconstructs the ASN.1 EC signature back into the `r||s` form Azure Key Vault's `Verify` expects, but used `r` for **both** halves (`r||r`) instead of `r||s`. As a result **every** ECDSA verification through this provider fails, e.g.:

```
could not verify envelope: accepted signatures do not match threshold, Found: 0, Expected 1
```

Before:

```go
rBytes := r.Bytes()
sBytes := r.Bytes()   // BUG: should be s.Bytes()
sigBytes = make([]byte, 0, len(rBytes)+len(sBytes))
sigBytes = append(sigBytes, rBytes...)
sigBytes = append(sigBytes, sBytes...)
```

After:

```go
ecPub, ok := pub.(*ecdsa.PublicKey)
// ...
keyBytes := (ecPub.Curve.Params().BitSize + 7) / 8
sigBytes = make([]byte, 2*keyBytes)
r.FillBytes(sigBytes[:keyBytes])
s.FillBytes(sigBytes[keyBytes:])
```

This fixes both the `r/s` ordering and the fixed-width padding (`big.Int.Bytes()` strips leading zeros, producing an incorrectly sized `r||s` that Azure also rejects).

**Scope:** affects all EC keys (P-256/P-384/P-521); RSA is unaffected. Affects any consumer that verifies via the KMS `SignerVerifier` (e.g. Sigstore bundle verification), not those that fetch the public key and verify locally.

**Regression:** introduced by #2355 while gating the EC (un)wrapping behind `signatureRequiresECWrapping`. Before #2355 (v1.10.5) it was correct `r||s`; after #2355 (v1.10.9 / current main) it is `r||r`.

**How reviewers can test:** run the new round-trip test, which uses a `kvClient` mock that mimics Azure Key Vault with real crypto (the existing `testKVClient` is a no-op `Verify` and cannot catch this):

```
$ go test ./...
ok  github.com/sigstore/sigstore/pkg/signature/kms/azure

$ go test -v -run RoundTrip ./...
--- PASS: TestVerifySignatureRoundTrip
    --- PASS: TestVerifySignatureRoundTrip/EC-P256
    --- PASS: TestVerifySignatureRoundTrip/EC-P384
    --- PASS: TestVerifySignatureRoundTrip/EC-P521
    --- PASS: TestVerifySignatureRoundTrip/RSA-2048
    --- PASS: TestVerifySignatureRoundTrip/RSA-3072
    --- PASS: TestVerifySignatureRoundTrip/RSA-4096
```

#### Release Note

Fixed Azure KMS ECDSA signature verification, which always failed because the raw signature was reconstructed as `r||r` instead of `r||s` and was not left-padded to the curve's byte width.

#### Documentation

NONE

### DIFF
--- a/pkg/signature/kms/azure/signer.go
+++ b/pkg/signature/kms/azure/signer.go
@@ -18,6 +18,7 @@ package azure
 import (
 	"context"
 	"crypto"
+	"crypto/ecdsa"
 	"errors"
 	"fmt"
 	"io"
@@ -181,12 +182,24 @@ func (a *SignerVerifier) VerifySignature(sig, message io.Reader, opts ...signatu
 			return errors.New("parsing signature")
 		}
 
-		rBytes := r.Bytes()
-		sBytes := r.Bytes()
+		// Azure Key Vault expects the raw signature as r||s, with each
+		// coordinate left-padded to the curve's byte size. big.Int.Bytes()
+		// strips leading zero bytes, which would produce an incorrectly sized
+		// signature (and thus a spurious verification failure) whenever r or s
+		// has a high zero byte, so pad explicitly using the key's curve size.
+		pub, err := a.client.public(a.defaultCtx)
+		if err != nil {
+			return fmt.Errorf("getting public key: %w", err)
+		}
+		ecPub, ok := pub.(*ecdsa.PublicKey)
+		if !ok {
+			return fmt.Errorf("expected ECDSA public key, got %T", pub)
+		}
+		keyBytes := (ecPub.Curve.Params().BitSize + 7) / 8
 
-		sigBytes = make([]byte, 0, len(rBytes)+len(sBytes))
-		sigBytes = append(sigBytes, rBytes...)
-		sigBytes = append(sigBytes, sBytes...)
+		sigBytes = make([]byte, 2*keyBytes)
+		r.FillBytes(sigBytes[:keyBytes])
+		s.FillBytes(sigBytes[keyBytes:])
 	}
 
 	return a.client.verify(a.defaultCtx, sigBytes, digest)
--- a/pkg/signature/kms/azure/signer_roundtrip_test.go
+++ b/pkg/signature/kms/azure/signer_roundtrip_test.go
@@ -1,0 +1,264 @@
+//
+// Copyright 2026 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azure
+
+import (
+	"bytes"
+	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/asn1"
+	"math/big"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/azkeys"
+	"github.com/jellydator/ttlcache/v3"
+)
+
+// These tests exercise the full SignMessage -> VerifySignature round-trip using
+// *real* crypto, via a kvClient mock that mimics Azure Key Vault: for EC keys it
+// returns/consumes a fixed-width r||s signature, and for RSA keys a PKCS#1 v1.5
+// signature. The existing testKVClient in client_test.go is a no-op (Verify
+// always succeeds), so it cannot catch signature (un)wrapping bugs. These tests
+// can, and guard against regressions like an incorrect ASN.1 -> r||s conversion.
+
+func hashForDigestLen(n int) crypto.Hash {
+	switch n {
+	case 48:
+		return crypto.SHA384
+	case 64:
+		return crypto.SHA512
+	default:
+		return crypto.SHA256
+	}
+}
+
+func azCurveName(c elliptic.Curve) azkeys.CurveName {
+	switch c {
+	case elliptic.P384():
+		return azkeys.CurveNameP384
+	case elliptic.P521():
+		return azkeys.CurveNameP521
+	default:
+		return azkeys.CurveNameP256
+	}
+}
+
+func signVerifyKeyOps() []*azkeys.KeyOperation {
+	return []*azkeys.KeyOperation{
+		to.Ptr(azkeys.KeyOperationSign),
+		to.Ptr(azkeys.KeyOperationVerify),
+	}
+}
+
+// realCryptoKVClient performs actual ECDSA/RSA signing and verification,
+// mirroring Azure Key Vault's raw signature formats.
+type realCryptoKVClient struct {
+	testKVClient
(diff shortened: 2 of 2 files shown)


ducet8:
Data point in support of merging this: the **padding half** of this fix resolves a long-standing intermittent production failure against real Azure Key Vault, not just the recent `r||r` regression from #2409. On pre-v1.10.9 releases (correct `r||s` ordering, no padding), ~1/128 verifies of a valid signature fail with `no matching signatures: failed vault verification` — whenever `r` or `s` has a leading zero byte, `big.Int.Bytes()` yields 31 bytes and AKV receives a 63-byte ES256 signature. We confirmed this cryptographically: the failing signatures verify fine offline with the exported public key, and each has a 31-byte component; passing ones are all 32+32. Full write-up: https://github.com/sigstore/cosign/issues/3709#issuecomment-5542394350. This has been biting cosign users intermittently since at least 2024 via sigstore/cosign#3709.

bobcallaway (dismissed):

bobcallaway:
@hugolevino can you fix the `gofmt` issue here and we'll get this merged/released? thanks!

bobcallaway (approved):

hugolevino:
> @hugolevino can you fix the `gofmt` issue here and we'll get this merged/released? thanks!

@bobcallaway it was CRLF line endings on the test file. Thanks!

PR #2417, opened 2026-09-09
adjust mldsa error handling

loosebazooka:
Addresses review comments on ML-DSA support from #2353 that were not in #2416.

- Consolidate ML-DSA uninitialized key panic probes into a single canonical function `cryptoutils.ValidateMLDSAPublicKey`, eliminating redundant ad-hoc `recover()` blocks across `pkg/signature` and `pkg/cryptoutils`.
- Add explicit documentation on `ValidateMLDSAPublicKey` explaining the deliberate panic probe on `.Parameters()` to catch uninitialized keys where internal parameters are nil.
- Remove blanket `recover()` from `cryptoutils.SKID` and scope ML-DSA key validation explicitly.
- Restore deprecation notice on `ValidatePubKey`.
- Update `AlgorithmDetails.checkKey` to return an error on typed nil and uninitialized ML-DSA keys so callers receive clear diagnostics instead of generic mismatch errors.
- Ensure symmetric nil message and option handling between `Sign` and `SignMessage` in `pkg/signature/mldsa.go`.
- Add comprehensive tests for `SKID`, `EqualKeys` (broken first and second arguments, separate representations), `ValidateMLDSAPublicKey`, `checkKey`, `GetDefaultPublicKeyDetails`, and `SignMessage`/`Sign`/`VerifySignature` option guards.

### DIFF
--- a/pkg/cryptoutils/publickey.go
+++ b/pkg/cryptoutils/publickey.go
@@ -87,6 +87,11 @@ func MarshalPublicKeyToPEM(pub crypto.PublicKey) ([]byte, error) {
 // subjectPublicKey (excluding the tag, length, and number of unused bits).
 // https://tools.ietf.org/html/rfc5280#section-4.2.1.2
 func SKID(pub crypto.PublicKey) ([]byte, error) {
+	if mldsaKey, ok := pub.(*mldsa.PublicKey); ok {
+		if _, err := ValidateMLDSAPublicKey(mldsaKey); err != nil {
+			return nil, err
+		}
+	}
 	derPubBytes, err := x509.MarshalPKIXPublicKey(pub)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/cryptoutils/publickey.go
```diff
@@ -116,6 +121,14 @@ func EqualKeys(first, second crypto.PublicKey) error {
...
 	case *mldsa.PublicKey:
+		if err := ValidateMLDSAPublicKey(pub); err != nil {
+			return err
+		}
+		if secondKey, ok := second.(*mldsa.PublicKey); ok {
+			if err := ValidateMLDSAPublicKey(secondKey); err != nil {
+				return err
+			}
+		}
 		if !pub.Equal(second) {
```
bobcallaway, line 127: doesnt this equality infer that the secondKey is also valid?

jku (approved):
Yeah this looks good to me
* I'm still not 100% sure the invalid key checks are needed but it feels better
* the stricter sign() argument handling is good, docstrings are better: this should make the API much harder to misuse accidentally

bobcallaway (approved):

pkg/cryptoutils/publickey.go
```diff
@@ -125,6 +133,24 @@ func EqualKeys(first, second crypto.PublicKey) error {
 	return nil
 }
 
+// ValidateMLDSAPublicKey checks that the ML-DSA public key is non-nil and properly initialized,
+// returning its parameters.
+// In the Go standard library, calling methods such as Parameters() on an uninitialized
+// &mldsa.PublicKey{} panics because its internal parameters are uninitialized. This function
```
Hayden-IO, line 139: Should this be provided as feedback to the std lib maintainers to fail gracefully?

loosebazooka: probably it's a bit of a weird behavior.

PR #2415, opened 2026-09-08
Bump go to 1.27 and fix linter and api issues

loosebazooka:
Just a more up-to-date version of #2394 with some fixes and rebased on main

#### Release Note
Update go to 1.27

PR #2431, opened 2026-09-17
Restore Go 1.26 support by gating ML-DSA behind go1.27 build tags

### LINKED ISSUE #2430: go 1.27.0 directive in v1.10.10 forces all consumers onto Go 1.27
**Description**

#2415 bumped the `go` directive of every module to `1.27.0`. As far as I can tell the only code that needs Go 1.27 is the ML-DSA support from #2416 and #2417, which imports `crypto/mldsa`. The rest of the tree builds and passes tests on Go 1.26 unchanged.

Because the `go` directive is a hard minimum for module consumers, v1.10.10 cannot be vendored by any project that still supports Go 1.26, which is a currently supported Go release:

```
$ go version
go version go1.26.8 linux/arm64
$ go build ./...
go: go.mod requires go >= 1.27.0 (running go 1.26.8; GOTOOLCHAIN=local)
```

podman-container-tools/container-libs (the image, storage and common libraries behind podman, buildah, skopeo and CRI-O) builds on the two currently supported Go releases, so the Renovate update to v1.10.10 was blocked for that reason: https://github.com/podman-container-tools/container-libs/pull/1166#issuecomment-5712213475. It stays on v1.10.9 until Go 1.28 ships and 1.26 goes EOL. Distributions that ship the previous Go toolset hit the same wall.

Proposal: restore `go 1.26.0` and gate the ML-DSA code behind `//go:build go1.27`, with `!go1.27` stubs so that ML-DSA keys degrade to the existing "unsupported key type" errors and the `ML_DSA_*` registry entries are absent on older toolchains. On Go 1.27 and later behaviour and the exported API are unchanged.

I have a PR ready that does this and is verified on both Go 1.26 and 1.27.

**Version**

sigstore/sigstore v1.10.10 (main at 4f4c3f8). Reproduced with Go 1.26.8 linux/arm64.

harche:
Resolves #2430

#### Summary

#2415 raised the `go` directive to 1.27.0 because the ML-DSA support from #2416 imports `crypto/mldsa`, so v1.10.10 cannot be used by any consumer still building with Go 1.26, a supported release. See #2430, the downstream-impact discussion on #2415 (https://github.com/sigstore/sigstore/pull/2415#issuecomment-5626813605), and the blocked update in container-libs (https://github.com/podman-container-tools/container-libs/pull/1166#issuecomment-5712213475).

Restore `go 1.26.0` in all modules and gate every `crypto/mldsa` reference behind `//go:build go1.27`, with `!go1.27` stubs. On Go 1.27+ behaviour and the exported API are unchanged. On Go 1.26 the ML-DSA types are absent, the `ML_DSA_*` algorithms are not registered, and ML-DSA keys fall through to the existing "unsupported key type" errors. The ML-DSA tests move to `go1.27`-tagged files with no assertions dropped.

CI runs the unit-test matrix on 1.26.x and 1.27.x; the 1.26 leg is what stops a future change from raising the floor again. Lint and CodeQL pin Go 1.27.x so the gated files are still analysed.

Tested with `go mod tidy`, build, test, `go test -race` and golangci-lint v2.13 on both Go 1.26.8 and 1.27.1.

#### Release Note

Restored support for Go 1.26. ML-DSA support is only compiled with Go 1.27 or later.

#### Documentation

README updated in this PR; no docs.sigstore.dev change needed.


.github/workflows/codeql.yml
```diff
@@ -44,15 +44,12 @@ jobs:
...
-        run: |
-          awk -F'[ .]' '/^go /{print "version=" $2 "." $3; exit}' go.mod >>"${GITHUB_OUTPUT}"
-
+      # Scan with the newest supported Go so that files gated behind
+      # "//go:build go1.27" (ML-DSA support) are included in the analysis.
       - name: Set up Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: "${{ steps.go_version.outputs.version }}"
+          go-version: 1.27.x
```
cpanato:
```suggestion
          go-version: '1.27'
```

.github/workflows/verify.yml
```diff
@@ -72,9 +72,11 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+      # Lint with the newest supported Go so that files gated behind
+      # "//go:build go1.27" (ML-DSA support) are included in the analysis.
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version-file: ${{ matrix.module }}/go.mod
+          go-version: 1.27.x
```
cpanato: same here

harche:
A note on the lifecycle of this change: once Go 1.28 is released and 1.26 reaches EOL, the `go` directive can move to 1.27.0 and the gating becomes inert (the `!go1.27` files are never compiled once the floor is 1.27). I'll send a follow-up PR at that point that removes the `mldsa_unsupported.go` stubs and the `!go1.27` test, drops the `//go:build go1.27` constraints, and moves the CI matrix and the lint/CodeQL pins forward, so the build tags don't outlive their purpose.

.github/workflows/codeql.yml
harche: Thanks, done.

.github/workflows/verify.yml
harche: Thanks, updated.

pkg/cryptoutils/mldsa.go
```diff
@@ -0,0 +1,74 @@
...
+// This file is only compiled with Go 1.27 or later; see mldsa_unsupported.go
+// for the behavior on older toolchains.
+
+// ValidateMLDSAPublicKey checks that the ML-DSA public key is non-nil and properly initialized,
+// returning its parameters.
+// In the Go standard library, calling methods such as Parameters() on an uninitialized
+// &mldsa.PublicKey{} panics because its internal parameters are uninitialized. This function
+// performs a deliberate probe to catch such panics safely and return a descriptive error.
+func ValidateMLDSAPublicKey(pub *mldsa.PublicKey) (params mldsa.Parameters, err error) {
+	if pub == nil {
```
rphillips, line 37: I'm not a huge fan of nil checks, unless nil actually means something.

harche: This check predates this PR (came in with #2417); this change just moves it into the build-tagged file. Without it the recover() still catches the nil deref, but the error message becomes a runtime-error string, and there are existing tests in cryptoutils and signature that assert the "must not be nil" message for a typed-nil *mldsa.PublicKey. Happy to drop it in a follow-up if you'd prefer, but I'd rather keep this PR a pure move.

loosebazooka:
woops, okay taking a look.

harche:
> woops, okay taking a look.

Thanks.

pkg/cryptoutils/mldsa.go
```diff
@@ -0,0 +1,74 @@
...
+			err = fmt.Errorf("invalid ML-DSA public key: %v", r)
+		}
+	}()
+	// Deliberate panic probe: calling Parameters() panics if pub is &mldsa.PublicKey{}
+	return pub.Parameters(), nil
+}
+
+// checkMLDSAPublicKey reports whether pub is an ML-DSA public key and, if so,
+// whether it is valid.
+func checkMLDSAPublicKey(pub crypto.PublicKey) (isMLDSA bool, err error) {
```
loosebazooka: This looks a bit weird, but being unexported, it's probably fine.

pkg/cryptoutils/mldsa.go
```diff
@@ -0,0 +1,74 @@
...
+		return false, nil
+	}
+	if _, err := ValidateMLDSAPublicKey(pub); err != nil {
+		return true, err
+	}
+	if !pub.Equal(second) {
+		return true, errors.New(genErrMsg(first, second, "mldsa"))
+	}
+	return true, nil
+}
```
loosebazooka: I feel like equal should still follow the contract that pub.equal uses (err or non-error)... 

It does require that it is gated by the use of checkMLDSA publicKey, but that's how those switch-case statements sort of look like anyway where we would use equals.

not necessarily this, but something along the lines of:
```suggestion
func equalMLDSAKeys(first, second crypto.PublicKey) (err error) {
	pub, ok := first.(*mldsa.PublicKey)
	if !ok {
		return errors.New("not a mdlsa key or wahtever")
	}
	if _, err := ValidateMLDSAPublicKey(pub); err != nil {
		return err
	}
	if !pub.Equal(second) {
		return errors.New(genErrMsg(first, second, "mldsa"))
	}
	return nil
}
```

pkg/cryptoutils/publickey.go
```diff
@@ -120,37 +117,15 @@ func EqualKeys(first, second crypto.PublicKey) error {
 		if !pub.Equal(second) {
 			return errors.New(genErrMsg(first, second, "ed25519"))
 		}
-	case *mldsa.PublicKey:
-		if _, err := ValidateMLDSAPublicKey(pub); err != nil {
+	default:
+		if isMLDSA, err := equalMLDSAKeys(first, second); isMLDSA {
```
loosebazooka: As mentioned above, this flow looks strange, seems like it could look like a continuation of the switch/case by doing:

```
if checkMLDSA:
  then check equals
```

loosebazooka (commented):
I think the equals behavior is weird, but other than that it looks fine

pkg/cryptoutils/mldsa.go
harche: Yeah, left this one as-is. It's the one place that does the type assertion + validation, so `EqualKeys`, `SKID`, and `ValidatePubKey` can all just gate on it.

pkg/cryptoutils/mldsa.go
harche: Went with your version. It returns a plain `error` now and assumes the caller already gated on `checkMLDSAPublicKey`. Kept the `ValidateMLDSAPublicKey` call inside so it's still safe if someone calls it directly. The `!go1.27` stub returns an error instead of `nil`, unreachable in practice but nicer than a silent pass. Tested on 1.26.x and 1.27.x.

pkg/cryptoutils/publickey.go
harche: Restructured like you described: `checkMLDSAPublicKey(first)` first, bail on its error, otherwise `equalMLDSAKeys(first, second)`. Everything else still hits "unsupported key type".

pkg/cryptoutils/mldsa.go
loosebazooka: actually can we just make this "isMLDSAPublicKey" and ignore the validate -- simplying returning true or false? I believe it will make a lot of the code easier to follow. Might require some minor adjustments at call sites.

harche: Done. `checkMLDSAPublicKey` is gone, replaced by `isMLDSAPublicKey` returning a bool and a separate `validateMLDSAPublicKey`.

Couldn't drop the validate step entirely though. `x509.MarshalPKIXPublicKey` calls `pub.Parameters()` unconditionally for ML-DSA keys ([x509.go#L120-L121](https://github.com/golang/go/blob/go1.27.1/src/crypto/x509/x509.go#L120-L121)), and `Parameters()` panics on a zero-value key because the embedded fips140 key has no params set ([mldsa.go#L100-L110](https://github.com/golang/go/blob/go1.27.1/src/crypto/internal/fips140/mldsa/mldsa.go#L100-L110)). A typed nil `*mldsa.PublicKey` nil-derefs on the same path. So `SKID` and `ValidatePubKey` still call validate after the type check, and `EqualKeys` just gates and hands off to `equalMLDSAKeys`, which validates internally before `Equal`.

Quick repro on 1.27.1:
```go
x509.MarshalPKIXPublicKey((*mldsa.PublicKey)(nil)) // panic: nil pointer dereference
x509.MarshalPKIXPublicKey(&mldsa.PublicKey{})      // panic: mldsa: internal error: unknown parameters
```

pkg/cryptoutils/mldsa.go
```diff
@@ -0,0 +1,80 @@
...
+// ML-DSA support depends on crypto/mldsa, which was introduced in Go 1.27.
+// This file is only compiled with Go 1.27 or later; see mldsa_unsupported.go
+// for the behavior on older toolchains.
+
+// ValidateMLDSAPublicKey checks that the ML-DSA public key is non-nil and properly initialized,
+// returning its parameters.
+// In the Go standard library, calling methods such as Parameters() on an uninitialized
+// &mldsa.PublicKey{} panics because its internal parameters are uninitialized. This function
+// performs a deliberate probe to catch such panics safely and return a descriptive error.
+func ValidateMLDSAPublicKey(pub *mldsa.PublicKey) (params mldsa.Parameters, err error) {
```
loosebazooka: It seems like this should just be called "MLDSAPublicKeyParameters" or something
1. that's the only way we seem to be using it
2. it's sort of clashing with lowercase "validateMLDSAPublicKey"
3. since we're 0.10.x, we have no stability guarantees on api/exported fucntion (we don't need to preserve this)

harche: Happy to rename it, it's a small mechanical change.

> since we're 0.10.x, we have no stability guarantees on api/exported fucntion (we don't need to preserve this)

Understood. One heads up anyway: go-tuf master already imports this by name against v1.10.10 ([keys.go#L153](https://github.com/theupdateframework/go-tuf/blob/master/metadata/keys.go#L153)), so they'd need a one-line fix when they bump. If that's fine with you I'll go ahead with `MLDSAPublicKeyParameters`.

loosebazooka: Yeah that's fine

loosebazooka: They are using it for parameters anyway

harche: Renamed to `MLDSAPublicKeyParameters`.

Hayden-IO (changes requested):
Thanks for flagging this. However, I don't think we should do this. This is adding a lot of code complexity for something that within 6 months (come Go 1.28) will go away once Go 1.26 falls out of support. You also mentioned the other service dependencies - they will all need to be bumped to 1.27 anyways to use crypto/mldsa, so this library requiring 1.27 isn't a blocker.

I think the problem actually was that we shouldn't have bumped the Go version in a patch release, v1.10.10. This release could have been in a v1.11.x release. Dependents don't have to update to the latest minor version, they can hold off until they update to the latest Go version. Releases of this library are infrequent, and if there happens to be a critical bug, we can always release a patch fix for 1.10.x.

@loosebazooka If you agree, could we [retract](https://github.com/sigstore/sigstore/blob/main/go.mod#L39) v1.10.10, and tag HEAD with v1.11.0?

loosebazooka:
I seem to have gotten my version numbers mixed up. We can retract 1.10.10, that was definitely a mistake. Make a 1.11.0 first and then retract?

harche:
Thanks, that's fair, and retracting v1.10.10 in favour of v1.11.0 would definitely help with the patch-release surprise. Two things worth weighing before deciding though.

> Dependents don't have to update to the latest minor version, they can hold off until they update to the latest Go version.

The `go` directive propagates through MVS, so holding at v1.10.9 only works if nothing else in the graph moves. go-tuf master already requires v1.10.10 with `go 1.27.0`, and fulcio/rekor/cosign will follow. Once that happens, consumers on Go 1.26 can't bump any of those either, retraction or not. container-libs (image/storage/common behind podman, buildah, skopeo, CRI-O) has an explicit two-versions policy and their maintainers suggested exactly this build-tag approach ([comment](https://github.com/podman-container-tools/container-libs/pull/1166#issuecomment-5712213475)).

> if there happens to be a critical bug, we can always release a patch fix for 1.10.x.

v1.10.10 already carries the Azure KMS ECDSA fix (#2410), the retryable transport (#2396) and the ed25519 length check (#2378), which those consumers would be locked out of unless each gets backported.

> This is adding a lot of code complexity for something that within 6 months (come Go 1.28) will go away

The new code is two stub files (~90 lines); the rest is moving existing ML-DSA code into `go1.27`-tagged files. When 1.26 goes EOL, cleanup is deleting the stubs and dropping the tags, and I'm happy to own that PR.

If you'd still rather go the retract route I understand, just wanted the trade-off on the table.

Hayden-IO:
Do you want to also create a tag v1.10.11 for e8841f5753b45451f47b703793ab5c6fd1dd46c8, right before the 1.27 bump commit? That'll pick up the few improvements made recently.

harche:
Thanks, a v1.10.11 at e8841f5 is a nice improvement for anyone who just needs the recent fixes.

It doesn't cover what motivated this PR though, so let me spell that out. container-libs doesn't need ML-DSA to work on Go 1.26. What it needs is to be able to *require* a sigstore version that has ML-DSA without raising its own `go` directive, because their policy is to build on both supported Go releases ([comment](https://github.com/podman-container-tools/container-libs/pull/1166#issuecomment-5712213475)). With `go 1.27.0` in sigstore's go.mod that's impossible: `go mod tidy` forces container-libs to 1.27 the moment it requires v1.11.0.

With the tags in this PR, container-libs requires v1.11.x and keeps `go 1.26.0`. Built on 1.26 it gets the stubs, same behaviour as today. Built on 1.27, which is what CRI-O would do, the same code gets real ML-DSA. That's the whole point: one sigstore version that both sides of the ecosystem can depend on for the next five months.

If the concern is carrying the stubs on main until 1.26 goes EOL, I'm happy to own the removal PR the day that happens.

mtrmac:
> (What we _would_ value is having a sigstore/sigstore branch which supports Go 1.26 as long as that language version is supported, with security backports, if any. But we can have that discussion if a need for such a branch arises.)

… and we would be contributing to that upstream work, of course.

Hayden-IO:
We've cut v1.10.11 where the min Go version is still 1.25.0, and v1.11.0 which is the same as v1.10.10 with the min Go version at 1.27.

This repo is fairly stable with infrequent updates. If backporting does become a burden or there is a rise in security reports (historically, we have had very very few in this repo), we can look at maintaining a separate branch.

harche:
@Hayden-IO thanks for taking the time on this, and for cutting v1.10.11. That unblocks the fixes side for Go 1.26 consumers, which is a good outcome.

